PR #94, opened 2021-10-31
Clean up package.json

### LINKED ISSUE #87: package.json needs a thorough cleanup
There are some issues with the `package.json` file:

1. There are packages like `babel`, `express`, `copy-webpack-plugin`, etc. that are listed under `dependencies`, but should be under `devDependencies` (I think).  `webpack` may be smart enough to not include these unnecessary modules, but it's better to have a clear distinction between the two lists of dependencies. 
2. There are some packages, like `brotli-webpack-plugin`, that don't seem to be used anymore and can be removed.
3. Many listed packages are 1 or more major versions behind.
4. `npm audit` lists 46 vulnerabilities, which require updating major versions to fix.
5. A number of the scripts use `babel-node`, which isn't installed with the repo, and reference a `tools/run` directory or script that's also not part of the repo.
6. Some of the scripts reference `ngrok` and `mocha`, which aren't listed in `package.json`.
7. Some of the scripts also assume `yarn` is installed.  I'm not sure what the best practices are there, but I feel like I usually only see `npm run` in `package.json` scripts.

fwextensions:
- Fix #87.
- Move development libs to devDependencies.
- Remove lots of unused libs.
- Update many libs to latest minor version.
- Update scripts to have consistent names and not use `yarn`.
- Remove scripts that used the `tools/run` path.
- Have `start:dev` start the webpack dev server and serve the files from `localhost:3000`.
- Get lint:css working.
- Fix test:coverage so it opens the HTML report by adding open-cli.
- Make 16.x the required Node version.

The script organization is based on personal preference, but happy to change it back.  There are still 46+ vulnerabilities listed by `npm`, which will probably require updating to react 17 and webpack 5 to fix. 

To test this PR, you should wipe `node_modules` and do a fresh install, to make sure I didn't miss any needed packages.


zoobot (approved):
Nice css linting, does that run on webpack? I have changed this dockerfile but will do another PR.

PR #95, opened 2021-11-02
Build custom useQuery hooks in queries.js

fwextensions:
- Use custom `useQuery()` hooks in `Cities.js` and `Countries.js`.
- Build all the get network functions with a `createQueryFetch()` factory in `apiUtils.js`.
- Move zoom handling in Cities/Countries into the `useEffect()` so it doesn't constantly log exceptions about the cities and countries layers missing as you zoom in.
- Remove `src/Slideout/`, which is a duplicate of `src/components/Slideout/`.
- Remove `api/actionUtility.js`, `src/DOMUtils.js`, `src/queries.js` and `src/registerServiceWorker.js`, which aren't used.
- Remove `mapboxgl.accessToken = config.mapbox.key;` line in Cities and Countries, since `config.mapbox` doesn't have a key property.
- Throw error if fetch `response.ok` is false, so that react-query will retry the call.
- Don't include an unnecessary ? on the URL when there are no params passed to `getData()`.
- Remove `{ fetchPolicy: cache-first }` from queries.js.
- Remove `console.log()` when zooming.


zoobot (commented):
Nice api changes and custom hooks, its much more clear! Can you double check all components that use useQuery to see if the queries/paths are correct with the new changes?

fwextensions:
I've only used the custom hooks for Cities and Countries, which are both simple GETs that get the full list.  They seem to generate the URLs correctly, but I haven't seen the response because I haven't gotten the postgis database to return anything yet.  

I'm holding off on making custom hooks for useMutate until the DB is working so I can see the POST and PUT calls actually making changes in the local docker.

fwextensions:
I found a bug in the fetch call (which I think may already be in the existing code), so I need to fix that before merging.

PR #98, opened 2021-11-04
changed to yarn again, added docker python for python error

zoobot:


fwextensions:
What was the python error?  I don't have it installed and haven't had issues installing modules locally.

fwextensions:
Weird that it would be Mac-specific inside a container.  As for installing locally on macOS, have you wiped node_modules and yarn.lock?  There were a bunch of upgrades and changes in package.json.

PR #99, opened 2021-11-05
Use custom useQuery and useMutation hooks in TreeData and children

fwextensions:
- Support creating useMutation hooks that use either POST or PUT, and can pass in either one string or an array.
- Create useTreeDataMutation and useTreeHistoryMutation hooks.
- Fix createAPIFetch() to handle query and data being passed in an array, as they are currently when getPost() is used directly.
- Return early with null if there's no treeData, instead of having everything hang off of a check on idTree.
- Arrange props in consistent order.

**Question**: the mutation hooks aren't actually used in `<TreeData>`.  They're just passed as props to its children.  If the mutation hooks are now created in just one place (`api/queries.js`), would it be simpler to just have the child components import the hooks as needed?

### DIFF
--- a/client/src/api/apiUtils.js
+++ b/client/src/api/apiUtils.js
@@ -29,11 +29,15 @@ function createAPIFetch(method) {
         url += `?${paramString}`;
       }
     } else {
-      url = apiEndpoints[query];
+      const [api, body] = Array.isArray(query)
+        ? query
+        : [query, data];
+
+      url = apiEndpoints[api];
       Object.assign(options, defaultPostOptions);
 
-      if (data) {
-        options.body = JSON.stringify(data);
+      if (body) {
+        options.body = JSON.stringify(body);
       }
     }
 
--- a/client/src/api/queries.js
+++ b/client/src/api/queries.js
@@ -9,22 +9,33 @@ function createUseQuery(api, defaultData = {}) {
   };
 }
 
-function createUseMutation(...apis) {
+function createUseMutation(apiList, method = 'POST') {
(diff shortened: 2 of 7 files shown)


zoobot (commented):
It would be nice if thats the case that you can just import the custom hooks as needed in the child components. What to test when making this change is:
When you change health or do tree maintenance like watering, pruning, etc, does it show up in treehistory on the bottom of the treedata? If you can edit those all separately AND together and they all show up in treehistory, then you are good to go with having them import separately.

fwextensions:
I moved the custom hooks out of `TreeData` and into its children in #100. 

> When you change health or do tree maintenance like watering, pruning, etc, does it show up in treehistory on the bottom of the treedata? If you can edit those all separately AND together and they all show up in treehistory, then you are good to go with having them import separately.

I don't think changes in tree health are recorded in the history.  There seems to be just one current health value stored in the tree object.  But different maintenance actions can be added on different visits to the dialog, and they show up as a single history entry (one per day).  Seems to also only be one comment per day.

This is the history for one tree after doing maintenance actions across different invocations of the dialog:

image

fwextensions:
Changing tree health adds a larger custom marker with a different color, though it changes after a refresh:

image

image

Removed trees show up like this:

fwextensions:
> It looks like color change is buggy on production now.

To be clear, the color changes immediately to that red color, but after a refresh, that tree with poor health shows up as brown, which is presumably the correct color.

I can take a look at the edit flow after these PRs.  That link should probably be hidden in prod until it's fixed.

PR #100, opened 2021-11-07
Refactor TreeData 💯

fwextensions:
* Refactor `TreeData` children into multiple files.
* Move custom like and adoption hooks from `TreeAdoptionLike.js` to `queries.js`.
* Simplify creation of data tables in `<TreeLocation>`, `<TreeMoreInfo>` and `<TreeHeader>`.
* Enable pressing <kbd>esc</kbd> or clicking outside the tree editing modal to close it.
* Rename `treedata_utilities.js` to `treeDataUtils.js`.
* Don't export `putData`, as it's no longer used directly outside of queries`.js`.
* Support passing a default options object to `createUseQuery()`.

This should be merged after #99.

Should these commits be squashed, first?


zoobot (commented):
If it makes it more clear for the history and part of the same feature or bugfix, squashing is always nice. If the changes are unrelated, no need to squash. 

Are we re-using the TreeHeader in AddTree? We should prioritize YAGNI easy to read code over extensibility if it's not being re-used more than a few times since its open source and has random developers who just jump in with no context. 

Having all the sub components in separate files is great, nice cleanup job and will make it easier to work on in the long run!! Nice!

fwextensions:
There are TreeHeader and TreeInfo components in both `addtree/` and `treedata/`, but they appear unrelated.  I didn't add any of those.  I just added a simple component to spit out rows of key/value pairs instead of having lots of repeated code.

PR #101, opened 2021-11-08
Use custom hooks everywhere

fwextensions:
- Add useCreateTreeDataMutation() custom hook for adding a new tree.
- Add custom hooks for user data and treemap.
- Don't export getData() or postData() from queries.js, since they're no longer used anywhere else.
- Refactor UserProfile.
- Move style definitions outside of the components.
- Remove userProfile overhead.
- Fix onPageChange in UserTreeHistoryTable.js.
- Rename currentTreeId props in TreeData children to idTree.
- Make the prop names consistent, even if the API param names aren't.
- Fix the TreeMaintenance timestamp to use the ISO standard.
- Flip PUT vs. POST check in createUseMutation().
- Enable add tree dialog to be closed with esc or a click outside.

### DIFF
--- a/client/src/api/queries.js
+++ b/client/src/api/queries.js
@@ -10,13 +10,13 @@ function createUseQuery(api, defaultData = {}, defaultOptions = {}) {
   };
 }
 
-function createUseMutation(apiList, method = 'POST') {
+function createUseMutation(apiList, method) {
   const apis = Array.isArray(apiList)
     ? apiList
     : [apiList];
-  const apiCaller = method === 'POST'
-    ? postData
-    : putData;
+  const apiCaller = method === 'PUT'
+    ? putData
+    : postData;
 
   return function() {
     const queryClient = useQueryClient();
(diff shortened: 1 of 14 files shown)


zoobot (changes requested):
The custom hooks are awesome! Looks so much cleaner and its great how the api mutations are all in one place. 

Can you revert the id changes to 
client/src/pages/treedata/TreeCare.js
client/src/pages/treedata/TreeData.js 
client/src/pages/treedata/TreeData.js 
client/src/pages/treedata/TreeMaintenance.js

We should keep the name difference between the id and currentTreeId as currentTreeid is the selectedTreeId. If currentTreeId wasn't clear, I'm game to change it to selectedTreeId(though it will change once we decide on the name the idWtt/treehash should be). We are also going to change the id from idTree to a hash based off lng/lat, etc instead of using the db id.

fwextensions:
I've reverted the changes of `currentTreeId` to `idTree` that I made.  Note, though, that there was originally a mixture of names passed to the children of TreeData, so that mixture is now back to what it had been.  That may have come about from the POST calls seeming to expect an `idTree` param, while the GETs usually expected `currentTreeId`.

As far as names go, I lean towards just `id`, and letting the object you're accessing provide the context.  `tree.id` looks better to me than `tree.idTree`.  The API param names could then all migrated to `id`, or better yet, do something like `/api/tree/WTT123456` to access a tree with that ID.

PR #102, opened 2021-11-11
Refactor Mapper and show zoom/lat/long in URL hash

fwextensions:
- URLs to specific map areas can now be shared (and restored, once the app tracks the hash when navigating between pages).
- Move map layer creation into `TreeLayer` and `TreeCountLayer` components, grouped inside `MapLayers`.
- Check whether MapboxGL is supported and map is loaded before adding any layers or controls.
- Move `city=%` to default data for `useTreemapQuery`.
- Add an optional processor function param to `createUseQuery()`.
- Change `mapLoaded` to `isMapLoaded`.
- Remove `showTree` state and toggle the TreeData modal based on `currentTreeId`.
- Remove `mutate(user)` call from Mapper.
- Remove `City/`, `Cities/`, `Countries/` and `mapper/mapper_utilities.js`, which was empty.
- Pass center and setCenter to AddTree.
- Don't call `setFeatureState()` on hover, as it's not needed to change the cursor to a pointer.
- Don't cluster points on the treemap layer, as they then don't line up with the location data.
- Support looking for both `id` and `id_tree` IDs when clicking trees.

image


zoobot (approved):
This is a really nice refactor. I love being able to see all the layers in one file. I kinda laughed at passing down giant props like the entire layer but its actually pretty clear and readable. We probably could break the layer out into a function at some point.
Thanks for getting the lng/lat into the url! Awesome work!! I'll try to bring this to production in the next few days. I want to spend some time on actions and see if I can automate that devops more.

fwextensions:
I thought adding the hash to the URL would be easy, as there's a built-in option for it.  But then tree layer wouldn't display immediately when the page loaded at a zoomed in position.  That led me down a rabbit hole of figuring out how the map layers were generated and triggered.  They were spread across Cities, Countries, makeLayerGeo() and makeLayerTile(), with lots of duplicated code.  

I hear you about the props, but creating a layer requires a big blob of keys/values somewhere, and at least they're all in one place now for easy comparison (and sharing things like circle colors).  Just trying to keep my +/- ratio negative.  ;)

image

PR #104, opened 2021-11-15
Move specification of health value/color mappings to treeHealth.js

fwextensions:
- Fix the health color when a new marker is inserted after the health is changed.
- Fix the slider not showing the correct health name when the dialog opens.
- Scale tree dots as map zooms in or out.
- Use solid colors in slider track and generate the background gradient programmatically from the health values.
- Move `config.js` to `util/`, change `mapbox` to `mapboxAccessToken`, and export `env`.
- Move `saveTimer` constant to `util/constants.js`.
- Generate slider datalist from the color constants.
- Share circle paint styling between tree layers, with same scaling.
- Use `isTreeMaintainable()` in TreeCare to determine whether to show the maintenance component.
- Don't show maintenance component for stumps.
- Remove `setShowTree` props, which are no longer used.
- Remove `mapbox_utilities.js`, which is no longer used.

This PR moves the definition of health colors to one place, fixing a bug where different colors were used in different places.  It also enables scaling of the tree circles, so they don't overwhelm the map when zoomed out.  It looks like this was initially intended, but the scaling was based on properties that don't seem to be returned in the vector tiles. (The temporary dots added to the DOM are still bigger because of their stroke color.  Not sure of the history there.)

**Before:**
image

**After:**
image

It also changes the health slider background to have solid colors, which map to fill colors used for the dots, since there are only 7 discrete colors and not a smooth gradient.  The colors for vacant, dead, missing and stump health are also distinguished, instead of all being white.  The positioning of the thumb is a little wonky, as it's not centered on each color bar, but that could be fixed if needed.

**Before:**
image

**After:**
image

This also adds a top-level `util` directory and moves `config.js` there, and adds a `constants.js` file (though it just has one export right now).  Happy to move them someplace else if there's a preferred location.


zoobot (commented):
I like the tree circle size change! @fwextensions 

What is noopQuery for? 
For things like this, should we have comments in the code?

For  client/src/pages/treedata/TreeHealth.js, can you add comment with example of what the colors data will look like so if someone wants to edit it will be clear what they need without running it? 
Also comment what's in ${colors.join(', ')}? It MAY already be obvious in the util... but if the files are separate from where its called, add a brief comment so people don't have to hunt for it?

I think we should not change default colors until we have the color key merged that Seimei did. It will confuse the Tree planting/maintenance volunteer teams. 
Also we should probably discuss color defaults per country as different colors mean different things in each country. (Red and white?)
I will check Seimei's branch after this and see if I can get it merged. May need a refactor with all the existing refactor.

Is healthNum for SF coming from the SF trees csv? Where'd you see that field? 
We should totally do a mapping for that to health! Many cities may not have health, definitely prefer default green than default red for a tree map! I believe (Gordon), the Oakland city tree person, said that the tree inventory company defaulted most trees to fair condition instead of good. 

If this PR doesn't change the whole map to red or change the default colors, can you add the few comments and merge?

Thanks for more awesome code!

fwextensions:
> What is noopQuery for?

I moved the default value for the `useQuery` prop out into its definition, so it's not recreated every time the component renders.

> Also comment what's in ${colors.join(', ')}? 

Added comments.

> I think we should not change default colors until we have the color key merged that Seimei did. 

The colors that were changed were for the dead, vacant, missing and stump health states, which were all white.  They did have different stroke colors, but only when first edited.  After they're returned in a vector tile, they're all white dots.  Original colors:

```
        'circle-color': [
          'match',
          ['get', 'health'],
          'good', '#309000',
          'fair', '#889944',
          'poor', '#be9b7b',
          'dead', 'white',
          'vacant', 'white',
          'missing', 'white',
          'concrete', '#808080',
          'stump', 'white',
          '#309000',
        ],
```

I changed them to different colors mostly so the slider wouldn't have 4 white states, but those states are also distinguished on the map:

**Before:**

image

**After:**

image

> Is healthNum for SF coming from the SF trees csv? Where'd you see that field?

`healthNum` comes back from the `trees?currentTreeId=` call.  All the trees in Oakland seem to have a corresponding `health` string that matches the number, but none of the SF trees do (Alameda has the strings, but nearly all of the trees seem to be set to `"good"`).  Compare this [SF tree](http://localhost:3002/api/trees?currentTreeId=225216) to this [Oakland one](http://localhost:3002/api/trees?currentTreeId=138014).

I generated that all red map on Trello just by changing that last default color in the paint array above to red.  Since all the SF trees had `null` for `health`, they all got the default color.  I tried mapping to the `healthNum` value, but that doesn't seem to be returned with the vector tiles. 

> If this PR doesn't change the whole map to red or change the default colors, can you add the few comments and merge?

It doesn't change anything to red, but it does change the white health states mentioned above to distinct colors, and the look of the health slider.

fwextensions:
@zoobot The original colors in the "before" screenshots have been restored, though the dot scaling is still enabled.  It's easy to re-add new colors when they're decided on.

PR #105, opened 2021-11-17
[DNM?] Update to webpack 5

fwextensions:
- Update webpack.config.js to work with v5.
- Update loaders and plugins.
- Use assets for font packaging to fix parsing error
- Clean the output folder before building.
- Add --progress profile to the build:analyze script.
- Remove file-loader and url-loader, since they're no longer needed.
- Comment out the MiniCssExtractPlugin instance for .scss files, since it doesn't seem to work and breaks SpeedMeasurePlugin in webpack 5.

Updating to webpack 5 eliminated the high priority vulnerabilities, leaving 8 moderate ones.  But the initial dev build when running `start:dev` is about **6X slower** than webpack 4 (~65 seconds).  The production build seems faster, but I'm not sure why the startup for watching file changes is so much slower.  Maybe it's Windows only?


fwextensions:
Not sure if the readme is the right place for these transient instructions, since anyone new will get the latest packages. Maybe just a slack announcement? Could also add a check to webpack.config for awhile.

PR #106, opened 2021-11-18
Add a legend to filter types of trees on the map

fwextensions:
A `TreeLayer` component is created for each health value (good, fair, etc.), including empty values.  The `MapboxLegendControl` component then shows an option to show/hide each layer.

image

### To do:

- [X] Figure out how to style the legend icons.
- [x] Decide on colors for the various health states.
- [x] Add a layer for all the trees that have no health data (like in SF).
- [x] Figure out how to share the vector tile source across multiple layers.
- [x] Fix visible, disabled scrollbar on Windows.


fwextensions:
I've updated the colors based on the mapping suggested in the [comment above](https://github.com/waterthetrees/wtt_front/pull/106#issuecomment-973710071).  The exact color values are easy to tweak, if needed.  The circles are now opaque, which makes it easier to control their color.  There's a separate layer for trees with health no data, which covers nearly all of SF.  

image

This PR also uses the solid colors in the health slider, so it's clearer where the slider stops are and how they map to the tree colors. 

This has been rebased on #109, so that should be merged first.

image

PR #108, opened 2021-11-19
Zoobot/sidebar

zoobot:
I'd created a side bar with the previous filter we were trying.


fwextensions:
I've installed the branch, but I'm not sure I see what's different.  The Adopt slider is still modal, like in production, right?  Is there another side panel?

PR #109, opened 2021-11-21
Replace Bootstrap with Material UI

fwextensions:
- Remove bootstrap and reactstrap from package.json.
- Remove bootstrap.min.css import from Layout, saving about ~700KB in the bundle.
- Create ScrollableDialog component to replace bootstrap Modal.
- Don't show Edit button in TreeHeader when running in prod.
- Add the bootstrap styles that were actually used to app.scss and remove multiple imports of it, to avoid duplicated CSS.
- Remove unused scss files from styles/ and Layout.scss.
- Remove variable uses and unused .progressbar from TreeData.scss.
- Clean up index.html, removing commented out favicon links.
- Use event.currentTarget.name in TreeMaintenance, so that clicks on the button labels are handled correctly.
- Simplify button generation in TreeRemoval.
- Remove Page.js, which isn't used anywhere.
- Remove extraneous TreeData.scss imports.

Some of the button colors have changed a bit with MUI, but the functionality is the same.  Things can get cleaned up when the UI is moved to a side panel.

image

image


zoobot (changes requested):
This is a huge amount of work to get rid of Bootstrap! Great start!! 👍 👍 👍 👍 👍 👍 👍 
Can you have a go at putting it in a side panel before merging or make a few changes?

The Tree # at the top is not necessary/ won't be used by volunteer tree planters. 

Prod version with small margins has more real estate for content on mobile so looks better.

The button text and icons should stay on the buttons on all sizes. We should make sure to check on all mobile devices with Chrome dev tools as that's what most volunteers will be using out in the field.

Buttons look good in dark mode as the prune/other icons are light. If we go light buttons, we should change icons.

Can that acid pink button for delete be red or yellow instead?

Add a Tree title and tree icon should be on one line.
Private/Public should be on one line.

I think that's it.

Probably once its on the side, things will look different too so may be worth it to wait to merge this. 

Sorry to many change requests! :[  Awesome you got rid of Bootstrap!! WOOWOOOOOWOOO!!!

fwextensions:
I should merge PR #110 over this before making lots of changes to the styling, as that upgrades to the latest MUI v5.  That should be a more stable base for building up new styles.  The default icon button layout and color changes in v5 anyway, so that'll need to be updated.

image

fwextensions:
The dialogs now go full-screen on phone-sized pages.  Fixed the maintenance toggle button colors and layout.

image

image

PR #110, opened 2021-11-23
Upgrade to MUI v5, React 17, and latest react-query

fwextensions:
- Use standard variant on the FormControls that contain menus so the label text shows up in the right place with MUI v5.
- Upgrade to React 17.
- Upgrade to latest react-query.
- Remove @material-ui libraries.
- Add support for passing in dialog actions to ScrollableDialog.

MUI v5 is [the latest release](https://mui.com/blog/mui-core-v5/).  It revamps the styling and theming based on the emotion library, though it can still be styled with external CSS.

This should get merged after #109.


fwextensions:
These changes have been merged into #109.

PR #119, opened 2021-11-30
Show all trees in one menu

### LINKED ISSUE #12: Need edit TreeName button for when trees are replaced
On Tree info page, we need a small edit button so the Tree Name can be edited when the tree dies and is replaced but retains the same location/serial number.

fwextensions:
- Fix #12.
- Remove tree type radio button.
- Build a combined list of trees from SF, California and US food lists.
- Convert tree data files to JSON in `/data/json` and output built .js files to `client/src/data/dist`.
- Dedupe the trees and normalize all the common names to title case during the `build:data` step.
- Create `<TreeSelector>` to wrap common, scientific, genus fields into a component that's used in both AddTree and TreeData.
- Add `sortTreesBy.js` to sort the tree list by lowercase strings.
- Refactor the creation of the name autocomplete menus into `<FormAutocomplete>`.
- Create Form component to wrap `<form>` with `<FormProvider>`.  Use it in `TreeDataEdit.js` and `AddTreeModal.js`.
- Make Edit button in TreeData work and show it in prod again.
- Remove treeList.js, MuiAutoComplete.js and MuiRadioSelector.js, which are no longer used.
- Remove Mui* CSS classes that aren't being used.

This actually adds ~150K to the bundle, as the tree names are duplicated in a combined list in `data/dist/trees.js`.  But it sets things up to include the full data on just the pages that need it (`Data.js`), and adds a build step to dedupe and massage data combined from disparate sources.

When the common name of a tree is picked from the first autocomplete menu, the scientific and genus names are filled in automatically (including for the "Vacant Site" option).  This way, the user gets the right scientific/genus names by default, but can still edit them if needed.  Clearing the common name also clears the scientific/genus fields, though just editing it leaves them alone.

image

image


fwextensions:
I started this looking at making a component that could be shared between the add tree dialog and the edit flow from TreeData.  That's when I noticed the React errors with duplicate keys and the strange menu behaviors.  Seemed liked building a more streamlined data structure from the existing files would make things easier, and would also lead to a smaller bundle.

Agreed about the path.  I think the data should actually be behind a simple interface, so the components using it don't know/care whether it's coming from the server or is local, as it probably (?) makes sense for the server to eventually provide the tree list.  I think it's possible to wrap it in a simple `useQuery()`, which could eventually make a server call.

> Did you check that the data component still works with the changes and has all the fields?

Yes, the built files include just the fields the data page needs, and they work in AddTree as well.  The data format hasn't changed so far; just less of it.

zoobot (commented):
Can the data build with webpack? It'd be good to have less separate scripts/things that can break on deploy.

fwextensions:
Running any of the build scripts in package.json will build the data files.  I suppose we could make a custom loader to generate the files, but I've usually seen build tasks triggered by npm scripts or gulp or something.

fwextensions:
I've also updated this to increase the circle radius range to 12:1 - 24:16, so that the dots aren't too big when zoomed out, but get much larger when zoomed all the way in.  They're now bigger than the locator icon when zoomed all the way in, but don't overwhelm the map when zoomed out.

PR #137, opened 2021-12-06
Zoobot/fix/vacant site

zoobot:


client/src/pages/addtree/TreeInfo.js
```diff
@@ -12,21 +13,24 @@ import Widget from '../../components/Widget';
...
-  <MenuItem key="<1 inch" value="<1 inch">less than 1"</MenuItem>
-].concat(
-  [...Array(25).keys()]
-    .slice(1)
-    .concat(30, 36, 42, 48, 56)
-    .map(value => <MenuItem key={value} value={value}>{value}"</MenuItem>)
-);
+// Create an array of DBH/height menu items: [.25,.5,.75,1,2,3,4] inches.
+const treeSmallValues = [0.25, 0.50, 0.75];
+const treeLargeValues = [...Array(452).keys()];
```
fwextensions, line 18: Should the here range really be `[1..451]` for a newly planted tree?

client/src/pages/addtree/TreeInfo.js
```diff
@@ -12,21 +13,24 @@ import Widget from '../../components/Widget';
...
 // Create an array of tree health items, reversed so that "good" comes first.
 const healthItems = treeHealth.getNameValuePairs().reverse()
   .map(([name]) => <MenuItem key={name} value={name}>{name}</MenuItem>);
 
-export default function TreeInfo({ treeList, register, control, errors }) {
+export default function TreeInfo({
+  treeList, register, control, errors,
+}) {
   return (
     <Widget title="Tree Info" classes="treeinfo">
```
fwextensions, line 36: The `<MuiAutoComplete>` lines have been replaced with `<TreeSelector>` in #119.

client/src/pages/addtree/TreeInfo.js
zoobot: The trees we planted today were at least 3 years old. Many trees are old when they get planted. We watched a Disney land documentary and they basically trawled the new freeway getting build to grab the trees that were going to get cut down otherwise, they dug em up and put them at Disney land. I just added the largest in case someone wants to put it on the map. None street trees can be mapped so if someone's out in the woods, they can have at it.
[tree truck](https://5.imimg.com/data5/QT/EI/MY-5507310/tree-transplantation-service-500x500.jpg)

client/src/pages/addtree/TreeInfo.js
zoobot: I saw that after the fact. Merge conflicts, blah! Which one's easier to go in first?

client/src/pages/addtree/TreeInfo.js
zoobot: Planted date can be back dated... Should probably check if it goes back a few thousand years :D

fwextensions: I think I saw that documentary, about the Jungle Cruise? 

Google says the tallest tree in the world is 380ft, so 451ft is probably overkill...

Instead of a menu, just have a numeric field where people can enter the number directly?

client/src/pages/mapper/MapLayers.js
```diff
@@ -7,10 +7,18 @@ import { treeHealth } from '../../util/treeHealth';
 const circleRadius = {
   'circle-radius': [
     'interpolate', ['linear'], ['zoom'],
-    12, 1,
-    18, 6,
+    12, 3,
```
fwextensions, line 10: These radii make quite chunky circles, with a lot of overlap between trees.

client/src/pages/treedata/TreeHeader.js
```diff
@@ -21,13 +22,13 @@ export default function TreeHeader({
...
-          <h5>Height: {height}</h5>
-          <h5>DBH: {dbh}</h5>
+          {datePlanted && <h5>Planted: {planted}</h5>}
+          {height && <h5>Height: {height}</h5>}
+          {dbh && <h5>DBH: {dbh}</h5>}
         </div>
       )}
 
-      {env !== 'prod'
+      {env !== 'production'
```
fwextensions, line 31: This `env` string comes from `config.js`, which uses `prod`.  But this check is removed anyway in #119, since editing the tree type works there.

client/src/pages/addtree/TreeInfo.js
fwextensions: GH says there are no conflicts, so should be fine.

client/src/pages/treedata/TreeHeader.js
zoobot: thanks, thought I removed that.

client/src/pages/addtree/TreeInfo.js
zoobot: yeah Jungle Cruise! I thought I saw somewhere diameter was around there. Tried to shoehorn width and height together. Numeric field would probably be easier... dropdown's a pain on the phone.

client/src/pages/mapper/MapLayers.js
zoobot: The little dots are hard on to tap on the phone when in medium zoom. Not sure what a happy medium would be, agree it might be too fat when zoomed out, sort of want them larger when zoomed in. Definitely not perfect yet. I don't know what would be perfect. When zoomed out, fat dots looks like forest mode

fwextensions: I think keeping them small when zoomed out is fine, as there's no way to pick one at that scale with something as pudgy as a finger.  The zoom range used for the interpolation could be increased, so they're bigger when zoomed in more than 18.  I don't think it helps tappability if they're overlapping.  It may be possible to increase the tap target without increasing the visible pixel width.

There's also something weird going on next to the adopt and plant buttons.  For about 2/3 the width of the screen, dragging there scrolls the whole page rather than moving the map.  I think there's something transparent sitting over the map.

zoobot: I just fixed that 2/3 thing, the sidebar was too wide, and moved the buttons to the top because when in the field they kept going off the bottom of the screen when I used the phone zoom instead of the map zoom.

PR #146, opened 2021-12-12
Create TreeEditDialog to edit tree name and size

fwextensions:
- Create TreeNameAndSize to share fields between add and edit dialogs.
- Use numeric field controls to enter height and DBH.
- Create FormErrorMessage, FormDecimalField, FormTextField, FormCheckbox, FormSelect and FormTreeGroup components.
- Move Cancel/Add Tree buttons in AddTreeModal to dialog footer.
- Use MUI Alert component to render error messages.
- Use RHF context provider instead of passing control and errors down to children.
- Do form error checking on each change instead of waiting for submit.
- Fix rounded corners on groups in AddTreeModal.
- Pass pathname and hash to auth0 redirect, to try to send logged in user back to the same location (not currently working).
- Remove unused pages/error/, Widget.js, TreeDataEdit.js, ButtonsResult.js and AddTreeFields.js.

This PR moves the tree name and height/diameter editing into a modal dialog from the inline modal state in TreeData (the dialog is full-screen on a phone).  The name and size controls are shared with the AddTreeModal dialog, so they behave and look the same.  The size values are now entered in numeric fields instead of a menu.  I haven't been able to test it, but they should show the number keyboard on mobile. 

There are also some visual tweaks.  The error messages use the Alert component from MUI, but if we want to change their look, it's easy to update since it's just in one place now.  The groups in the Add Tree dialog have left-aligned white text.  The [FormTreeGroup.js](https://github.com/waterthetrees/wtt_front/blob/e30dfc42639f1d6bf497412da6a3167223db23e5/client/src/components/Form/FormTreeGroup.js) file is an example of using CSS-in-JS with MUI v5.  

**Desktop modal Edit dialog:**

image

**Mobile Edit dialog:**

image

**Add Tree dialog with new error messages and footer buttons:**

image


zoobot (approved):
Do you think TreeEditDialog is extensible enough to be able to use elsewhere as well, such as in the data table rows and columns? At some point, I'd like to have city data in there from like last 1 day/week/month of planting or something. @fwextensions 
Great Job! Huge amount of effort and work here!!

fwextensions:
> Do you think TreeEditDialog is extensible enough to be able to use elsewhere as well, such as in the data table rows and columns?

Not sure how it would be used on the `/data` page on the site, if that's what you're referring to.  Is it possible to query the backend for trees planted after a certain date?

fwextensions:
I originally had just the names in a `<TreeSelector>` component, but then added the size fields and made it `<TreeNameAndSize>`.  But it would be easy to bring that back if we wanted a component just for name picking.

PR #147, opened 2021-12-14
Add TreeRemoveDialog to handle modal remove flow

fwextensions:
- Create FormScrollableDialog.js to simplify dialogs.
- Make Dead the default reason for removal.
- Use FormScrollableDialog in TreeEditDialog.
- In edit dialog, allow names to be unchanged but replace checkbox checked and have the tree updated.
- Add FormScrollableDialog and FormRadioGroup to form/index.js.
- Remove unused CSS.

There's currently not an "are you sure?" confirmation after clicking Remove Tree, but we could add one.

**Desktop remove dialog:**

image

**Phone:**

image

### DIFF
--- a/client/src/components/Form/FormRadioGroup.js
+++ b/client/src/components/Form/FormRadioGroup.js
@@ -1,0 +1,42 @@
+import React from 'react';
+import {
+  FormControl,
+  FormLabel,
+  RadioGroup,
+} from '@mui/material';
+import { Controller, useFormContext } from 'react-hook-form';
+import FormErrorMessage from './FormErrorMessage';
+
+export default function FormRadioGroup({
+  name, label, options = [], rules, sx, ...restProps
+}) {
+  const { control } = useFormContext();
+
+  // We have to wrap the Controller around the RadioGroup, not the FormControl, since the
+  // RadioGroup is what is updating the actual value.
+  return (
+    <>
+      <FormControl
+        component="fieldset"
+        fullWidth
+      >
+        <FormLabel component="legend">{label}</FormLabel>
+        <Controller
+          name={name}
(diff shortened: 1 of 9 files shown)


zoobot (approved):
This looks way better! :) 

It is nicely more succinct without all those reallyDelete hooks. 
I am not sure its easier reading but it is almost 11 so that's likely more my problem. 

What is this doing?  What functions are getting sent in from where?  Should we comment this for future 11pm end of day people like me?
const isFunction = (value) => typeof value === 'function';

fwextensions:
`isFunction()` is just there to not repeat `typeof` checks when making sure the optional `on*` props in `FormScrollableDialog` are functions before calling them.  They could also be set to a noop function in the props destructuring, which could avoid that check.  

But now that I'm explaining it, the `onConfirm()` and `onCancel()` props really aren't optional, as the caller needs some way of knowing which button was clicked.  So maybe it's fine if they error if they're not functions.

fwextensions:
Yeah, just if they're not passed in.  `onError()` is optional, as we seem to just be logging that to the console when it happens.  So defensively checking the type of the other props isn't really necessary.

fwextensions:
One question: what does the `no water` reason mean?  Isn't the goal to get people to water the trees if they're not getting enough as it is? Or is that the tree died from not enough water?

fwextensions:
So maybe the removal options should be something like:

- Dead from lack of water
- Dead from insects or disease
- Dead from unknown causes
- Other

I'll remove those function checks just to avoid confusion.  Seems like people are moving to TS to validate props instead of PropTypes.

fwextensions:
Resolves #150 

Added new reasons:

image

PR #148, opened 2021-12-14
Update react-hook-form to 7.22.0

fwextensions:
- Move owner radio buttons to TreeAddress, since it's more related to the location than the planter.
- Rename TreeRemoveDialog to TreeRemovalDialog.
- Change `as` props to `render` props.
- Remove control props, as they're unnecessary when the Controller is in a FormContextProvider.
- Move restProps to the rendered element, as the Controller no longer forwards the extra props.
- Update watch() calls.
- Extract errors from formState, since it's moved.
- Pass menu items and radio buttons as children of Select and FormRadioGroup, instead of an options prop.
- Report errors immediately in TreeEditDialog and TreeRemovalDialog.
- Add FormRadio component for FormRadioGroup children.
- No margin on the last FormTreeGroup.
- Ignore react-hook-form in the source-map-loader, since it throws errors.

PR #147 should be merged first. 

image

### DIFF
--- a/client/src/components/Form/FormAutocomplete.js
+++ b/client/src/components/Form/FormAutocomplete.js
@@ -4,28 +4,25 @@ import {
   TextField,
   Autocomplete,
 } from '@mui/material';
-import { Controller, useFormContext } from 'react-hook-form';
+import { Controller } from 'react-hook-form';
 import FormErrorMessage from './FormErrorMessage';
 
 export default function FormAutocomplete({
-  name, label, options = [], rules,
+  name, label, options = [], rules, ...restProps
 }) {
-  const { control } = useFormContext();
-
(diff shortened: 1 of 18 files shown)


zoobot (approved):
What errors are they? 

// These modules seem to cause errors with this loader.
          exclude: /mapbox-gl-legend|react-hook-form/,

fwextensions:
> What errors are they?

That loader seems to be there to look for source maps associated with libs in `node_modules`.  I'm unclear if it's really necessary, but it was there when I started looking at webpack.  I made it so it only runs in dev.  It seems like reasonable looking source is shown in devtools for those modules, so the loader may not be needed at all.  Haven't really looked into it.

PR #153, opened 2021-12-18
Update to react-router v6 and retain current page when logging in/out

### LINKED ISSUE #149: zoom level/coordinates not retained after login
Can we retain them after login so it doesn't zoom people all the way back out?

### LINKED ISSUE #152: update react router to v6 to latest
cleanup router legacy code sprawl

fwextensions:
- Resolve #149.
- Resolve #152.
- Update to react-router v6.1.1.
- When getting redirected back to / from the Auth0 login flow, navigate to the /go page with the path + hash stored in returnTo, so that the hash from Auth0 gets replaced with the previous one.
- On logout, return to the current page and hash.
- When user accesses a feature that requires login, don't show the feature while the login page is loading.
- Comment out source-map-loader to avoid errors while building dev.
- Move all auth-related files to components/Auth/.
- Move App.js to components/App/ and all routes to <App>.
- Refactor AuthButton to return one component that handles both the logged in and out states.
- Rename protected-route.js to RequireAuth.js.
- Rename auth/auth0-provider-with-history.js to Auth/Auth0ProviderWithRedirect.js.
- Rename client.js to index.js and limit it to mounting <App />.
- Remove unused Layout.js, *-button.js files and user.js.
- Update the entry name to index.js in webpack.config.js.
- Fix NotFound page routing.
- Don't show a disabled scrollbar in map legend.
- Center 404 message vertically on the page.
- Get rid of that broken button in the map attribution.

When logging out of, say, the `/about` page, it's supposed to redirect you back to that page after logging out via the Auth0 site.  Currently, that's showing an error before redirecting back, saying the URL isn't listed in "Allowed logout URLs".  I'm not sure if that's because it's running on localhost, http, or that only `/` is allowed.  We can either add the other page URLs on the Auth0 management page, or the redirect can just go back to the homepage. 

Gone!

image


zoobot (approved):
This made me laugh because the 1st 5 or 6 pages were deleted. 
Once this is on dev server, can you test to make sure everything works again?
Hopefully upping the ram will have fixed the action.

fwextensions:
Most of those deletes were renames or moves, but that didn't seem to make it through the commits.

PR #161, opened 2021-12-22
errors cannot render while render other component

zoobot:


fwextensions:
The tree ID calculation really seems like it should be done on the server.  That way, the client can just treat it as an opaque value and the server doesn't have to trust the client's calculation. 

How long will the ID be?  It would be nice to support URLs like waterthetrees.com/tree/<treeID> to zoom and center the map on a particular tree with the inspector dialog open.

PR #162, opened 2021-12-23
Split the code into chunks

fwextensions:
- Load the /data page's code on demand, so that the user doesn't incur the cost of the full tree data when they're just viewing the map.
- Re-enable source-map-loader, while ignoring react-router.
- Clean up the Loading element and add a visible icon.
- Don't load app.scss in Sidebar.scss, and let the App component include it.

This increases the usage ratio of the JS loaded on the main page from 53% to 91%.

### DIFF
--- a/client/src/components/App/App.js
+++ b/client/src/components/App/App.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Suspense } from 'react';
 import { BrowserRouter, Routes, Route } from 'react-router-dom';
 import { QueryClient, QueryClientProvider } from 'react-query';
 import { createTheme, ThemeProvider } from '@mui/material/styles';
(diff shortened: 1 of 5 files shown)


zoobot (approved):
We've had issues with Suspense on and off. We've pulled it out a number of times so please test thoroughly on dev! Let me know when you are done and I will bring current code to prod in the next few nights.
I like that you changed the loading image! Sorry to see those food trees go, good idea to just put in a filter instead of pull it out of the data, that way it future proofs it if we grab a new file or see more weirdo trees! 🍡 surprised there was no candy apple trees. Thanks for the changes!

fwextensions:
What kind of suspense issues have you seen?  This is the first time I've used it, but I haven't noticed any problems so far.  It's only used for the fallback when loading the /data page.  That could also be replaced with the `loadable-components` library.

fwextensions:
I've tried dev on an iPhone and iPad with iOS 15.2, which is hardly comprehensive, but I'm not sure why mobile would be different from desktop for this.  Is there an old branch with the suspense use on it?  

When I first started working on this, package.json included react 17 but react-dom 16, so possibly that had something to do with the bugs.  They're both now at 17.0.2.

PR #163, opened 2021-12-25
Remove all usages of MUI v4 makeStyles

fwextensions:
- Remove @mui/styles from package.json, since we're already including @emotion as the MUI styling engine.
- Refactor Slideout and Data.
- Pass slideout trigger button in as a prop instead of passing classname props in.
- Make the /data URL in the menu lowercase, like the others.
- Import the tree icon in Loading so the URL doesn't change depending on which module it's imported in.
- Make Plant and Adopt button text white, so they're easier to read.
- Add padding to adoption directions in slideout.
- Remove unused CSS.

This saves about 150K from the vendor bundle.


zoobot (changes requested):
I got an error/page crash on the data page when running this and changing between the tree dropdown selector. will post to slack

fwextensions:
I've pushed a fix for that crash.

PR #168, opened 2021-12-30
Create custom TreeLayerLegend to control the layers

fwextensions:
- Create MapboxLegend class to handle adding a React component as a map control.
- Remove mapbox-gl-legend library, saving ~450K from the bundle.
- Support collapsing and expanding the legend.
- Collapse the legend by default.
- Show hover state on each control.
- Fix default layer visibility being undefined when the map first loads.
- Add colored circular checkboxes for each layer.

The `mapbox-gl-legend` library was including the `legend-symbol` library, which was huge.  So writing our own React version saves a lot of code.  The colored dots don't scale as the map zooms, and now serve as the checkboxes for the legend.  

The legend is shown at the bottom-right, and collapses to a small button:

image image

PR #167 should be merged first.

### DIFF
--- a/client/src/pages/mapper/MapboxLegend.js
+++ b/client/src/pages/mapper/MapboxLegend.js
@@ -1,0 +1,34 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+export class MapboxLegend {
+  container = null;
+  map = null;
+  type = null;
+  props = null;
+
+  constructor({ type, props = {} }) {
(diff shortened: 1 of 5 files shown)


fwextensions:
BTW, the legend is now in the bottom-right, because I figured out why it wasn't showing there before (paren in the wrong place), but it can be moved back to the top-right or wherever.  We can also collapse it by default on small screens, if desired.

PR #179, opened 2022-01-10
Zoobot/feature/geohash2

zoobot:


fwextensions:
Haven't taken a close look yet, but noticed a couple `console.log()` calls.

client/src/pages/treedata/TreeCare.js
```diff
@@ -7,7 +7,7 @@ import { treeHealth } from '@/util/treeHealth';
 export default function TreeCare({
   currentTreeId, common, health,
 }) {
-  const { data: treeHistory } = useTreeHistoryQuery({ currentTreeId });
+  const { data: treeHistory } = useTreeHistoryQuery({ id: currentTreeId });
```
fwextensions, line 10: One thing we could do is change the custom hooks to take just the `id` as a param, instead of a whole param object.  That way, the param name wouldn't be repeated everywhere, and the caller wouldn't have to care what it is.

client/src/pages/treedata/TreeData.js
```diff
@@ -65,15 +65,15 @@ const TreeContent = ({ currentTreeId, map }) => {
 
       {showEditDialog && (
         <TreeEditDialog
-          idTree={idTree}
+          currentTreeId={id}
```
fwextensions, line 68: Consistency FTW!

client/src/pages/treedata/TreeRemoval.js
```diff
@@ -48,7 +48,7 @@ export default function TreeRemoval({ idTree, common, notes }) {
       const newNote = `${common} was removed by ${user.nickname} ${today} - "${comment}"`;
 
       sendTreeData.notes = (notes && notes !== newNote)
-        ? `${notes ? notes + '\n' : ''}${newNote}`
+        ? `${notes ? `${notes}\n` : ''}${newNote}`
```
fwextensions, line 51: Airbnb linting noise...

client/src/pages/treedata/TreeCare.js
zoobot: You mean just have id instead of currentTreeId? I thought about changing them all to id but having currentTreeId is helpful to see where its getting passed around vs where id is coming in from api. Or do you mean just a global custom hook?

zoobot: I want to test the vector tiles so just trying to get the id's into dev so I can move on to that as a separate branch. I've made the "required" id changes, without much refactor. 
Would be nice to have a custom hook for id, we should add that in as a separate issue though?

client/src/pages/treedata/TreeRemoval.js
zoobot: I have mine set to lint on save...

client/src/pages/treedata/TreeCare.js
fwextensions: I just meant passing a bare number instead of an object, like:

> const { data: treeHistory } = useTreeHistoryQuery(currentTreeId);

That way, you don't need to care about how that is sent to the server.  For multiple params, it would need to be an array, like `useTreeLikesQuery([idTree, user.email])`.  Of course, that gets more complicated with multiple params, as then the order matters, so not sure how helpful it would be.

zoobot: The object is more clear for full stack developers, though it's a bit more code.

PR #194, opened 2022-01-13
Add side panel to show selected tree details or new tree controls

fwextensions:
- Add `/map` route to access the new behavior.
- Add MapLayout component to manage the map and tree details.
- Add Map component to render the mapbox and controls.
- Show a selection state for the current tree on a separate layer.
- Update Address section in new tree panel as the marker is dragged around.
- Save changed notes to the backend after a short delay or on loss of focus.
- Create useNewTree() hook to handle NewTreePanel interactions via a context.
- Add MapboxControlPortal to wrap a Mapbox control created with React, and use it with the Plant and Adopt buttons.
- Show the adoption directions in a dialog instead of a drawer from the left or inline in the details drawer.
- Add ErrorBoundary.js to show an alert at the bottom of the screen.  Closing the alert will reload the page.
- Set react-query to not refresh the queries when the window is refocused.
- Refactor createUseQuery() to take an options param and enable a query to modify the request before it's passed to useQuery(), so that it can filter out certain requests to avoid server errors, like /trees?currentTreeId=null returning a 500.
- Create Like, Liked, Adopt, Adopted checkboxes with appropriate fill colors.
- Move maintenance buttons and notes field into a dialog.
- Move Edit link to same row as like and adopt.
- Add Map.css to include mapbox style overrides
- Add `npm run docker` script to start docker-compose.
- Add use-debounce and react-error-boundary packages.

image

image

### Things to do and see:

- Open http://localhost:3000/map to see the new features, as the current behavior remains on the standard routes.
- Click a tree to see its details in a drawer that slides in from the right.  Click an empty part of the map to clear the selection and close the drawer.  The map will shift if necessary to keep the selected tree visible after the drawer opens.  
- Click Plant and then click the marker to open the new tree flow in the drawer.  This acts modally, with Add/Cancel buttons at the bottom, and disables tree selection while it's open.   
- On mobile, the details and new tree panels open in a modal dialog, as before.
- The Notes field is saved automatically after the user stops typing for a second.
- The Maintain Tree button opens a modal dialog with the action buttons and a field for maintenance notes.
- The Plant and Adopt buttons are rendered as map controls, as in the other two corners, but this could be changed.
- The icons on the Maintain Tree and Remove buttons were just whatever was closest in the MUI icon gallery.  

### To do, probably in a later PR:

- [ ] When moving the mouse quickly, it's possible to move it over the tooltip.
- [ ] Disable tooltips while dragging the new tree marker.
- [ ] The add tree panel snaps open without animating.


fwextensions:
Use useImperativeHandle() to get ref to map from child to MapLayout? Or maybe useRef is enough.

zoobot (changes requested):
AWESOME!!  First of all, The new sidebar looks WAY better!! I love it! Thanks for all your hard work on this!!

Please change for consistency and for ease of integrating new vector tiles as I already have a branch open on that:
Can you add your new slide out features, any new map code into the mapper folder component and the new tree into addtree keeping the same folder structure/file naming that currently exists in development?

Documentation for newbs:
Can you leave comments in queries.js so its easy to understand immediately?

Awesome:
Thanks for adding error boundaries!! We've needed that for a while!

Awesome:
Thanks for fixing the tooltip styling!

Opinion:
We'd previously decided to stick with BEM styling, can we have that with styled-components? 
Personally I prefer having separate stylesheets than styled-components as I like less dom clutter. Worth more discussion so we are consistent? I know a lot of people enjoy styled-components and most pros and cons internet articles love styled components. Maybe I should keep mind open or just don’t get it?

Please change for safety reasons:
Please move the remove button to the bottom of the tree inspector so it’s not as easily visible. We don’t want people removing trees by accident just because they click on it instead of tree maintenance.

Please change for accessibility:
Please make Tree history text larger. Lotta reading glasses people are tree planters! I think I may need them.

Opinion:
I think controls should be on left under the add tree so it doesn’t slide on every tree selection. Sliding is fancy and slick but possibly distracting? What do you think?  Did you already try it over on left? It looks good on right when sliders not open.

Bug:
When adding a tree, and then deciding to unclick plant button, the new tree window should close.  

AWESOME Changes you’ve been contributing!! :) YEEHAW and HAZZAH!!

fwextensions:
> Please change for consistency and for ease of integrating new vector tiles as I already have a branch open on that:
> Can you add your new slide out features, any new map code into the mapper folder component and the new tree into addtree keeping the same folder structure/file naming that currently exists in development?

Once we're happy with the drawer stuff, I was going to remove the `addtree`, `mapper` and `treedata` directories, so that everything under `pages/` actually maps to a page with that name on the site.  We could merge the vector tile changes first, and I could copy over any needed changes.

> Can you leave comments in queries.js so its easy to understand immediately?

I'll add some JSDoc comments on the create functions.  Or is it more about react-query in general?

> Thanks for adding error boundaries!! We've needed that for a while!

Right now it only wraps the side panel, but I can wrap the whole app instead.

> Personally I prefer having separate stylesheets than styled-components as I like less dom clutter. Worth more discussion so we are consistent? I know a lot of people enjoy styled-components and most pros and cons internet articles love styled components.

By dom clutter, do you mean the inline MUI `sx` props?  I've mainly used use them in one-liners to adjust spacing.  The nice thing about them is you can do something like `{ my: 2 }` to set top and bottom margins to twice the standard spacing offset specified in the theme.  That way, you don't end up with lots of random spacing values and can adjust them all in one place. 

If it's the `` styled('div')`...` ``-type calls, they can be extracted into their own files and then imported where needed. 

> Please move the remove button to the bottom of the tree inspector so it’s not as easily visible. 

Is it the color or position of the button, or both?  It's not as eye-catching if it's not red:

image

Layout-wise, it seemed nice to have the buttons that affected the whole tree clustered at the top, and it avoided a blank space to the right of the Maintain button.  But it's easy to move to the bottom.  

Actually, I wonder if it should be part of the maintenance dialog?  Removing a tree is sort of its Final Maintenance.

> Please make Tree history text larger. Lotta reading glasses people are tree planters! I think I may need them.

Yeah, the theme sets it to `.875rem`, and the body text is set to only `12px`, so it's fairly small.

> I think controls should be on left under the add tree so it doesn’t slide on every tree selection. Sliding is fancy and slick but possibly distracting? What do you think? Did you already try it over on left? It looks good on right when sliders not open.

The drawer should only slide open when changing from no selection --> selection.  Changing from one selected tree to another should leave the drawer open.  

The smooth transition came from a MUI example I just copied and pasted.  It's also possible to have the drawer just snap open with no animation, which might be less distracting.  

> When adding a tree, and then deciding to unclick plant button, the new tree window should close.  

Easy fix.

fwextensions:
I just added a commit that removes the drawer animation, so you can see if that's better.  The Remove Tree button is moved to the bottom of the panel.  And the History and Info tables are set to `1rem` instead of `.875rem`.  I also set the MUI theme to use the font stack specified in `App.css` instead of using Roboto, so that may have a small effect on the font size, too.

fwextensions:
In terms of the directories, the new code doesn't quite map into the current structure, as it sometimes appears in a drawer vs. a dialog.  I was also trying to follow some best-practices articles for file organization, keeping the top-level directories in pages/ mapping to actual routes, vs. UI elements that would only ever appear within one route.  But happy to chat more about it.

fwextensions:
I slightly prefer the animation, but thought the controls on the right might be less distracting if the drawer just snapped in.

I quickly hacked smaller buttons in with everything on the left.  Might work, though I don't know if the icons would be immediately clear.

Speaking of icons, the adopt icon is different in the drawer vs. the top-left button.  Should they be the same? 

image

image

fwextensions:
I actually modified an MUI icon to make an SVG version of the droplet, as the CSS version was difficult to resize.  So it could be easily swapped out for another icon.  

Speaking of icons, are the original source file for the favicon, maintenance icons, etc. anywhere?  Probably would be worth checking them in somewhere in case changes or new versions needed to be made.

Definitely agree about not wanting to stomp on files others are working on, and that's part of why I limited the changes to a new path.  I took a quick look at the `feature/new-vector-tiles` branch.  This PR shouldn't interfere with those changes, and I think they can be merged in either order.  I'm happy to migrate over any required changes once the new vector tiles are ready.   

For the new vector source, is the plan to keep the existing source in parallel with the new one, or just switch everything?  If the former, then I think the legend would need to be reworked, since there'd be 2 layers with good trees, 2 with fair, etc.

fwextensions:
I just added some fixes, so I think the PR is ready to go.

PR #196, opened 2022-01-13
Add `eslint-import-resolver-alias` so lint understands the `@` alias

fwextensions:
Added the webpack alias path to `.eslintrc.js`.


fwextensions:
@zoobot merging simple fix for lint error.

PR #202, opened 2022-02-03
Add a location tracking toggle to the planting marker UI

fwextensions:
- Create the planting marker using React components via a portal.
- Add crosshairs to the marker so you can see exactly where the tree will go.
- Add tracking state to NewTree to control whether geolocation is being constantly updated.
- Add a Plant button to marker and change NewTreePanel to say `Plant New Tree`.
- Add MapboxMarkerPortal.js to wrap the React marker and manage adding it to the Mapbox map.
- Add TooltipLeft/Right components.
- Disable tree popups while the marker is being dragged.
- Hide any popup when the map selection is disabled.
- Move map to keep new tree location visible after drawer opens.
- Memoize the useNewTree action dispatchers, so they aren't created every time the custom hook was called, triggering a re-render.

image image

This PR adds some new functionality to the plant tree marker:

- A crosshair where the tree will be added.
- An explicit *Plant* button on the marker opens the new tree panel.  (Clicking on the circle bit doesn't do anything.)
- A toggle button to move the marker to the user's current location and track it, or turn the tracking off.

The location tracking is on by default for mobile users, but desktop users can also enable it if desired.  If the geolocation API is not available, on either desktop or mobile, the marker defaults to the center of the screen.  

Instead of forcing the Mapbox geolocate control to provide the user's location, it uses the geolocation API directly.  The location tracking toggle state is retained across hiding/showing the plant marker.  It switches off if the user drags the marker, so that manual moves don't fight with updates from the tracking.

The icon for the location toggle may be somewhat confusing.  It's sort of a crosshair, but it's not where the tree will go, and the marker now has its own crosshair.  It *is* the icon that Mapbox and Google Maps use on the web to indicate location tracking, though iOS and Android OS-level icons are different.  But maybe a "point of interest" icon would be clearer.  Something like:

image image

The trigger for opening the new tree panel could also be an icon button instead of a text button.


fwextensions:
Here's an alternative layout for the marker.  I think the location pin is less confusing than the crosshair icon.  Not sure the "plant" icon works at that size.  May just need to be a plus icon.

image

fwextensions:
Having a consistent disc behind the icons makes things look more cohesive.  I also created a different crosshair, so it doesn't look so plus-y.

image image

Also added custom tooltips for the buttons, instead of the OS ones.

image

fwextensions:
This branch has been rebased on development.  It includes the design for the marker from the first comment, not the later one with two circular buttons.

PR #204, opened 2022-02-03
Filter `American black bear` from the tree list

fwextensions:
Also clean up some airbnb complaints.

(An American black bear in a tree still doesn't make it a tree.)

image


zoobot (approved):
bye bear

PR #211, opened 2022-02-27
Clean up from side panel PRs

fwextensions:
- Convert MapboxMarkerPortal to functional component.
- Lazy-load the /map route.
- Wrap the app in an ErrorBoundary component.
- Fade in the loading page content so there's not a flicker of the tree image before the /map route loads.
- Make handling of edited notes work like the health slider and fix a bug where edited notes didn't show after a reload.
- Create useIsMobile hook using media query.
- Update docs and vars in queries.js.
- Make the marker buttons darker and add a white border to the outside it shows up better on a complicated background.
- Suppress mouse events on the buttons so the map doesn't hear them and show tooltips.
- Show an orange dot in the middle of the crosshairs.
- Clean up TreeDetailsPanel and CoreData.
- Export a marker offset value from Marker.js so NewTree can adjust where the tree is planted based on the marker location.
- Add postinstall script to run build:data so that the JS files are created in src/data/dist when the repo is created.
- Add build timestamp to bottom of index.html.

The tweaked planting marker looks like this:

image

### DIFF
--- a/client/src/api/queries.js
+++ b/client/src/api/queries.js
@@ -13,14 +13,16 @@ import { getData, postData, putData } from './apiUtils';
  * `useQuery` function with each call to the hook.
  * @param {function(string, object, function, object)} [options.preProcessor] - An optional function
  * that can modify the custom hook's parameters before they're passed to `useQuery`.
- * @param {function(*)} [options.postProcessor] - An optional function that can modify the results of the
- * API call before they're returned by the hook.
- * @returns {object} - See the [`useQuery`](https://react-query.tanstack.com/reference/useQuery)
+ * @param {function(*)} [options.postProcessor] - An optional function that can modify the results
+ * of the API call before they're returned by the hook.
+ * @returns {function([object], [object])} - A function that wraps a call to `useQuery`.  The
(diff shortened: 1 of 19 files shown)


zoobot (commented):
TreeEditDialogue is more descriptive than CoreData, 
Would keep the names consistent onCancel, onConfirm, onError /handleError, handleConfirm, handleCancel.

Could be consistent and descriptive with naming with notes, newValue could be newNote, trimmedNote? setValue?

fwextensions:
I don't love "CoreData" as a name, either, but that dialog isn't what I'd consider a full "edit tree" dialog.  It lets you edit only some of the core fields shown at the top of the side panel (like scientific name and DBH, but not date planted), and definitely not everything that's in the add dialog, like address or planter name.  So I as trying to be precise with the name and not overpromise what it was for. 

For "notes" vs. "value", I was trying to distinguish the current value shown in the control from the data stored on the server.  Same with the Health slider.  These controls have a `value` prop, so the local `value` state maps to that prop, while the `notes` or `healthNum` var refers to what's in the tree data sent from the server.

PR #217, opened 2022-03-11
Create React GeolocateControl component

fwextensions:
- Create UserLocationProvider and wrap MapLayout in it.
- Create useUserLocation hook to share context between GeolocateControl and NewTree, and centralize access to the browser geolocation API.
- When following the user, hide the geolocation marker and animate the planting marker.
- In the GeolocateControl, show a blue icon with a dot when tracking the location and not planting, or when the planting marker is following the user.  If tracking is on but following is off, remove the center dot.
- Default tracking on when tapping Plant on mobile.
- Make planting marker grey instead of blue and make crosshairs yellow.
- Add z-index to tooltips so they appear above the marker.
- Refactor marker code and files.

The new `GeolocateControl` appears in the same place in the top-right control group, but is built with React.  When turned on, it shows a pulsing dot where the user is:

image

If planting mode is then turned on, the marker replaces the disc and shows a pulsing animation itself, to indicate that the marker where the new tree will go is following the user's geolocation:

image 

The user can also pull the marker away from the user's current location and place it anywhere.  The user's location is still shown with the pulsing dot, while the marker is grey.  Tapping the geolocation button again will sync up the dot and marker.

image


zoobot (approved):
useReducer makes me regret removing redux!! and useMemo! 
I am merging this then going to re-organizing the code structure while I add in the id code, btw.

fwextensions:
I've read code that used redux, but never written any.  Reworking the new tree flow was the first time I've used reducers and contexts.  Hopefully it's a reasonable approach.

PR #234, opened 2022-04-13
set zoom levels across layers

zoobot:
The zoom levels were different for geoJSON and vector tiles so the tiles would disappear when zoomed in and though it started as a feature for adding trees, it just seemed like a bug.

### DIFF
--- a/client/src/pages/Map/Map.js
+++ b/client/src/pages/Map/Map.js
@@ -99,6 +99,8 @@ export default function Map({
         style: 'mapbox://styles/waterthetrees/ckyckqkqz8e4b14rn3rm1hh9k',
         center: [-122.34725, 37.7343787],
         zoom: 10,
+        maxZoom: 18.5,
+        minZoom: 2,
         // Pass true to update the browser URL hash with the current zoom and lat/long of the map.
         hash: true,
       });
--- a/client/src/pages/Map/MapLayers.js
+++ b/client/src/pages/Map/MapLayers.js
@@ -4,31 +4,17 @@ import { treeHealthUtil } from '@/util/treeHealthUtil';
 import TreeCountLayer from './TreeCountLayer';
 import TreeLayer from './TreeLayer';
 
-// TODO make the dots bigger on zoom out so you can see which cities have trees
-const circleRadius = {
-  'circle-radius': [
-    'interpolate', ['linear'], ['zoom'],
-    10, 2,
-    11, 3,
-    17, 9,
-  ],
-};
+const linearZoom = ['interpolate', ['linear'], ['zoom'],
+  5, 10,
+  7, 7,
+  10, 2,
+  18, 6,
+];
+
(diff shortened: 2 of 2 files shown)


client/src/pages/Map/MapLayers.js
```diff
@@ -4,31 +4,17 @@ import { treeHealthUtil } from '@/util/treeHealthUtil';
...
-const circleLayerZoomRange = {
-  type: 'circle',
-  minzoom: 5,
-  maxzoom: 18,
-};
+const linearZoom = ['interpolate', ['linear'], ['zoom'],
+  5, 10,
+  7, 7,
+  10, 2,
+  18, 6,
```
fwextensions, line 11: Why is the radius getting smaller and then bigger as the zoom goes up?  And why are the circles visible so far out?  Shouldn't just the city names be visible at that zoom level?

client/src/pages/Map/Map.js
```diff
@@ -99,6 +99,8 @@ export default function Map({
         style: 'mapbox://styles/waterthetrees/ckyckqkqz8e4b14rn3rm1hh9k',
         center: [-122.34725, 37.7343787],
         zoom: 10,
+        maxZoom: 18.5,
```
fwextensions, line 102: I found that when the map zoom was exactly the max zoom on the layers, the layers would disappear, so I maxed out the map zoom to .01 less than the layer max zoom.

client/src/pages/Map/MapLayers.js
```diff
@@ -126,16 +113,12 @@ export default function MapLayers({
...
             'circle-color': 'transparent',
-            ...circleRadius,
+            'circle-radius': linearZoom,
             'circle-stroke-color': 'white',
-            'circle-stroke-width': [
-              'interpolate', ['linear'], ['zoom'],
-              12, 2,
-              24, 12,
-            ],
+            'circle-stroke-width': linearZoom,
```
fwextensions, line 121: Making the stroke width the same as the radius seems like it'll make the selection border way too heavy.

client/src/pages/Map/MapLayers.js
zoobot: It looks nicer and easier to find the cities with trees when you see green  dots on zoom out.
The city names don't currently exist for all cities, just for ones in the db. We should remove that soon and have the layer created by another layer of vector tiles.

client/src/pages/Map/MapLayers.js
zoobot: Looks good to me, I don't see much difference if any to the way it was previously.

fwextensions:
Having the selection border the same as the radius seems way too chunky to me:

image

It would be good to get the city summaries back and show them when zoomed out.  It's kinda weird to have just one tree out of thousands selectable at that zoom level.  The cities could be selectable, showing counts, links and other data in the side panel.

PR #236, opened 2022-04-14
Update prettier rules

jazhen:
# Changelog
- Turned on formatOnSave in VS Code settings
- Added required packages "eslint-config-prettier" and "eslint-plugin-prettier" to integrate Prettier into ESLint
- Removed ESLint rules that dealt with code formatting. This is now only handled by the rules defined in the Prettier config file.
- Changed .prettierrc to .prettierrc.js and updated the formatting rules
  - I wanted be able to comment the Prettier config file with links to the documentation, so I chose a file extension that allowed that.
  
# View Proposed Changes
1. Checkout this branch `jazhen/235-update-prettier-rules`
2. Run the script: `npm run lint:fix:js`
3. View the changes. I used VS Code to easily view the diff side-by-side.


jazhen:
> Should we just deploy this lint in its own build next week onto prod after vector tiles and after all the details are hashed out?

Sure, let's just do it next week so we can discuss the the changes and address any specific concerns as a team.

Issue #237, opened 2022-04-14
Deal with vacant/possible tree/bad naming location site ids


zoobot:
Deal with vacant/possible tree/bad common name location site ids

In cases where common, scientific are malformed or missing or vacant site, how do we update the id?
Should we just delete old version and create new tree?

zoobot:
Option 1

1. Allow the id to be updated so it retains the history.
    a. update treedata, treehistory, users, treelikes, treeadoptions tables
    b. email city on edit
    c. Assume city does not update their download file
    4. Compare coordinates/planted dates and only show more recent one

Option 2

2. Allow multiples per coordinates
  a. copy history from old id but leave previous record intact
  b. check for most recent planted date per lng/lat
  b. email city data file of all monthly edits or combo of edits and their current data in their data format

Option 3

3. Redo the ids entirely without common and scientific
  a. Use only lng/lat/sourceId or even just a lng/lat geohash
  b. Deal with collisions by adding to tree count.

fwextensions:
As of #238, users will be able to remove trees but not re-add them in the same spot., until this gets resolved.

zoobot:
Option 3 is out, way too many collisions

3. Redo the ids entirely without common and scientific
    a. Use only lng/lat/sourceId or even just a lng/lat geohash
    b. Deal with collisions by adding to tree count.

zoobot:
Try option 4

4. Redo ids entirely without common and scientific- try without the geohash, try without the sourceid+geohash

zoobot:
Option 4 is a no go, too many collisions.

PR #238, opened 2022-04-14
deal with missing common

zoobot:


client/src/pages/Tree/treeutilities.js
```diff
@@ -11,3 +11,19 @@ export async function updateOrCreateTree(id, fullData, updatedData) {
...
+export const checkForMissingData = (currentTreeData) => [currentTreeData?.common,
+  currentTreeData?.scientific]
+  .some(undefRequiredField)
+    || ['vacant',
+      'vacant site',
+      'unsuitable site',
+      'asphalted well',
+      'planting site',
+      'other',
+      '#n/a', '::'].includes(currentTreeData?.common.toLowerCase());
```
fwextensions, line 26: Given that this is only used in one component, I don't think it needs to be in its own file, and the updateOrCreateTree() function above this should be deleted anyway.

More importantly, having `vacant` or `asphalted well` in the `common` field doesn't mean the entry is missing data.  You should be able to add a note or edit the common name and check the "New tree" option to indicate a new tree as planted in that location.  You just can't maintain or change the health of a vacant site.

zoobot: Agreed, I'll move it back, thanks

For the vacant and planting locations, those are actually very important as they are potential trees. 

Issue is we don't have it set up to deal with the id change that will happen when common changes. 
We will have to update the id or add a new tree with a different Id. Potential for multiple records with same lng lat.

I want to test a few things before allowing these to be saved.

Definitely want a way to add tree requests in these.

fwextensions: > Issue is we don't have it set up to deal with the id change that will happen when common changes.
We will have to update the id or add a new tree with a different Id. Potential for multiple records with same lng lat.

That's an issue for any tree, then.  Sounds like the Edit link should be removed for all trees until the ID handling is addressed.

zoobot: Its not a super trivial issue. Trees die often enough that the edit function is definitely a must have.

Option 1
1. Allow the id to be updated so it retains the history. 
  a. update treedata, treehistory, users, treelikes, treeadoptions tables
  b. email city on edit
  c. Assuming city does not update their download file, compare coordinates/planted dates and only show more recent one

Option 2
2. Allow multiples per coordinates
  a. copy history from old id but leave previous record intact
  b. check for most recent planted date per lng/lat
  b. email city data file of all monthly edits or combo of edits and their current data in their data format

Option 3
3 Redo the ids entirely without common and scientific
  a. Use only lng/lat/sourceId or even just a lng/lat geohash and deal with collisions by adding to tree count.

zoobot: Issue here  #237

client/src/pages/Tree/Tree.js
```diff
@@ -53,24 +57,26 @@ export default function Tree({
           <>
             <TreeHeader
               currentTreeData={currentTreeData}
-              vacant={vacant}
+              hasMissingData={hasUnfitData}
```
fwextensions, line 60: This prop is misnamed.  The `TreeHeader` is looking for `hasUnfitData`, so the `!hasUnfitData` check there returns true and the like/adopt/edit buttons are shown even when the `hasUnfitData()` call above returned true.

PR #245, opened 2022-04-22
Share trees by URL

jazhen:
When a tree is selected, we now add the tree's id to the URL as a hash parameter. As a result, we can save and share trees through their URL.

https://user-images.githubusercontent.com/6326660/164611502-a6fa0b1c-095c-4c19-aada-3f0db69d0a6a.mov


zoobot (approved):
This is Great! :)

PR #246, opened 2022-04-22
Update ESLint rules

jazhen:
- Removed Airbnb ESLint rules
- Added new ESLint plugins or configured them to actually be used. Some were installed but not used. 
- Removed custom rules that deal with formatting since we leave that up to Prettier instead.
- Updated versions for all ESLint packages


jazhen:
@zoobot I can merge all this and the other formatting related PRs. I wasn't sure if I had the ok yet.

PR #249, opened 2022-04-23
Share trees by URL w/ error handling

jazhen:
When a tree is selected, we now add the tree's id to the URL as a hash parameter. As a result, we can save and share trees through their URL.

https://user-images.githubusercontent.com/6326660/164865152-5b4a8acf-d121-4d0d-b49f-f4831ecbe579.mov

This now handles the case when the id in the hash param references a tree that doesn't exist in the database.

https://user-images.githubusercontent.com/6326660/164865429-15a530a3-9dbc-4ab7-90dc-9f50a54f257d.mov

### DIFF
--- a/client/src/pages/Map/Map.js
+++ b/client/src/pages/Map/Map.js
@@ -101,8 +101,8 @@ export default function Map({
         zoom: 10,
         maxZoom: 18.5,
         minZoom: 2,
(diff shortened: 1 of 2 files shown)


jazhen:
Hmm interesting, I didn't see that issue. Yeah, we can draft this until we figure it out.

PR #253, opened 2022-05-03
Fetch tree images from database tree data

jazhen:
I added the trees from the database to the data files as `databaseTrees.json`. The changes relate to fixing any issues that came up from doing this.

- Was getting a duplicate key warning and needed to add a filter to ignore duplicate "vacant site" instances 
- Extracted functions `normalizeNames` and `replaceIrregularNames` to handle names with accented characters or other weird names.
- After adding the database trees, the alphabetical sorting broke in the new tree dropdowns. Changed `createTreeSorter` to use [Intl.Collator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator), which allows for more powerful sorting. It handles accented characters, punctuation, and letter case.
Screen Shot 2022-05-02 at 9 49 12 PM

### DIFF
--- a/client/src/components/TreeUtil/TreeNameAndSize.js
+++ b/client/src/components/TreeUtil/TreeNameAndSize.js
@@ -10,7 +10,14 @@ function getNames(treeList, name, firstItem) {
   // Otherwise, React will complain about duplicate keys in the menu.  Sort the trees by the
   // different fields before extracting the strings, so that the sort function can access the
   // per-tree lowercase sort fields.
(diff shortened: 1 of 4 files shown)


zoobot (approved):
Nice, is this all the trees in the current db? 

Does int collator need to be installed as an npm package?

jazhen:
Yes, this is all the trees in the current db.

Intl.Collator is native to JavaScript, so it doesn't require any dependencies.

PR #261, opened 2022-07-02
globe view, header change to make globe view look better

zoobot:
globe view, header change to make globe view look better


jazhen:
Globe view is very cool! Only question is if the header is supposed not take up the full width anymore?

This was before
image

and this is the new change
image

Also, the alignment is slightly off. There's a black bar above "WATER THE TREES" and the hamburger menu.

PR #274, opened 2022-08-11
tree crashing because of Object.hasOwn in Safari/duckduckgo on mobile

zoobot:
Tree.js crashing because of Object.hasOwn in Safari/duckduckgo on mobile

changed it to Object.hasOwnProperty and is working now on production already.

Production has this branch checked out but once its merged, I will pull main back to prod


jazhen (commented):
Great catch!

jazhen (approved):

PR #289, opened 2022-09-14
fix favicon not displaying in browser tab

MattPereira:
I usually add `<link rel="icon" href="%PUBLIC_URL%/favicon.ico" />` to the `<head>` of the `index.html` but i couldn't get it to work with WTT.

I found an npm package called `react-favicon` and it works on my local machine.

Screenshot from 2022-09-14 15-17-13

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -81,6 +81,7 @@ module.exports = (env) => {
     },
     plugins: removeEmpty([
       new HtmlWebPackPlugin({
+        favicon: './client/public/assets/images/favicons/favicon.ico',
         template: './client/src/index.html',
         filename: './index.html',
         minify: false,
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "start": "node server/index.js localserver",
     "start:docker": "npm run watch & npm run start:nodemon",
     "start:nodemon": "nodemon server/index.js localserver",
-    "start:dev": "webpack serve --env development",
+    "start:dev": "webpack --env development && webpack serve --env development",
     "test": "echo \"No test specified\"",
     "test:coverage": "echo \"No test specified\" && exit 0 && jest --coverage && open-cli coverage/lcov-report/index.html",
     "test:watch": "echo \"No test specified\""


MattPereira:
Oh my bad. For some reason it doesn't show any favicon in production for
me.

I've tried 3 different browsers and 2 different machines (linux and mac)
and all I'm getting is the default globe icon.

[image: Screenshot from 2022-09-16 10-28-43.png]
[image: Screenshot from 2022-09-16 10-28-37.png]

MattPereira:
Ya I definitely wasn't a fan of loading the favicon into the App component
either, it was just the only way i could get it to work for this project.

Normally I just add

<link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico" />
<link rel="manifest" href="%PUBLIC_URL%/manifest.json" />

to the <head> of the index.html and it works.

I'm still trying to figure out why that doesn't work in this case.

jazhen:
@MattPereira - You should be able to add a new line in the webpack config as shown here: https://stackoverflow.com/a/54178489

Here's the file you need to modify: https://github.com/waterthetrees/wtt_front/blob/main/webpack.config.js#L83

MattPereira:
I implemented the changes to the `webpack.config.js` file as suggested but I still can't get the favicon to actually display on my local machine.

webpack.config.js
```diff
@@ -81,6 +81,7 @@ module.exports = (env) => {
     },
     plugins: removeEmpty([
       new HtmlWebPackPlugin({
+        favicon: './client/src/assets/images/favicons/favicon.ico',
```
jazhen, line 84: Can you change the favicon path to `./client/**public**/assets/images/favicons/favicon.ico'? It'll then use the _public_ folder that is output from webpack's bundling.

But before that we have to address building that _public_ folder. Did some research on webpack and realized that I wasn't actually bundling our app before running it. Don't think anyone else was either 🤷‍♂️.

So we need to add the bundling to our start command in `package.json`. There's multiple ways we could do this but let's keep it straightforward and change our `start:dev` script to `"webpack --env development && webpack serve --env development"`.

Then run `npm run start:dev` and the favicon should appear. You might need to clear cookies on `localhost:3000` as well.

jazhen (changes requested):

webpack.config.js
MattPereira: It worked! Thanks for doing all of the heavy lifting lol. I was off course trying to `npm run build:dev` to see if the webpack changes worked. I really appreciate you taking the time to both help and explain your logic! 

Screenshot from 2022-10-11 14-36-50

jazhen:
@MattPereira - If you have time and are interested in doing more research do you want to give an general overview of how our webpack config works during the Wednesday meeting, in addition to your changes in this PR? I would just focus on at least what's in the [entry](https://github.com/waterthetrees/wtt_front/blob/main/webpack.config.js#L20), [output](https://github.com/waterthetrees/wtt_front/blob/main/webpack.config.js#L28,L32), [module](https://github.com/waterthetrees/wtt_front/blob/main/webpack.config.js#L54-L81), [plugins](https://github.com/waterthetrees/wtt_front/blob/main/webpack.config.js#L82-L103) keys. You can also do more if you feel up to it. Good learning opportunity and I can try to help fill in the gaps if you get stuck.

jazhen (approved):

PR #293, opened 2022-09-29
stylelint added

zoobot:
Added stylelint as our css linter. 
swapped BEM__style for kabab-case.
Still needs a bit of work. I'll add an issue to either add linting to the PR workflow or as a precommit.

@fwextensions added lint scripts in the package.json which are nice and we should use them until we have linting automated.
He moved much of the code to styled components. We should move whatever's left in stylesheets to styled components if it makes sense.


Tijani-zainab (approved):
Great work!
will work on moving what's left off of the stylesheet to styled component and fix whatever could be broken

PR #297, opened 2022-10-01
did some code clean_up in some component

Tijani-zainab:
did some code cleanup in some component, getting rid of some annoying vscode errors

…still working on moving the rest of the stylesheet to styled component, pr in batch for easier review!

### DIFF
--- a/client/src/components/Footer/Footer.js
+++ b/client/src/components/Footer/Footer.js
@@ -15,4 +15,6 @@ const Footer = () => (
   </div>
 );
 
+// export default Footer;
+
 export { Footer };
--- a/client/src/components/Footer/Footer.js
+++ b/client/src/components/Footer/Footer.js
@@ -15,6 +15,4 @@ const Footer = () => (
   </div>
 );
 
-// export default Footer;
-
 export { Footer };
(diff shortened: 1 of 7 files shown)


client/src/components/Footer/Footer.js
```diff
@@ -15,4 +15,6 @@ const Footer = () => (
   </div>
 );
 
+// export default Footer;
+
```
jazhen, line 19: Can you please remove this commented out line?

jazhen (changes requested):
Great work! Can you just fix one minor issue?

client/src/components/Footer/Footer.js
Tijani-zainab: Thank you!
 @jazhen  I actually wrote that piece of code trying to get rid of the `index.js` file and export default directly from the `footer.js` file and fix all imports where necessary. for some reason that broke the app even after fixing imports in other places.
water

I commented it out to get back to it later.
find out what the error is, see if I can fix then get rid of the `index.js` file

Tijani-zainab: but wait, when you say remove you meant get rid of it ? 

looks like I might be unclear what you meant, if it's to get rid of it. will do that
but if it's to uncomment it, that might cause some issues.

jazhen: Yup, just meant to get rid of it. Just wanted to make sure we don't have unused or in progress code committed if possible. Thank you!

jazhen (approved):
Good job on your first commit @Tijani-zainab . Thanks!

Tijani-zainab:
Thank you! @jazhen

PR #303, opened 2022-10-07
302 fix all eslint issues auto fixable

Tijani-zainab:
Hi 👋🏽 @jazhen 

ran the command to fix all-eslint-issues. went through each changed file too to see everything was good.


jazhen (approved):
Thanks @Tijani-zainab

Issue #323, opened 2022-10-19
[bug]: Undefined links in Tree Details


Vinitvh:
Links for Tree Details section shows `undefined data` and `undefined info`

![tree](https://user-images.githubusercontent.com/42197888/196623827-b59f9bee-38f5-4378-8a84-c86f16cee00b.png)

zoobot:
If there is no data, those links should not show up at all

jazhen:
Good place to start looking: https://github.com/waterthetrees/wtt_front/blob/main/client/src/pages/Tree/TreeLinks.js#L12-L17

Vinitvh:
Can we add inline conditional check for label, which returns an empty string for null value?

anup-a:
@jazhen do you think this is fine https://github.com/waterthetrees/wtt_front/pull/343?

PR #343, opened 2022-10-26
fix: display blank string if city is not defined

anup-a:
Addresses https://github.com/waterthetrees/wtt_front/issues/323.

In some cases there are download links but cities aren't defined, which causes UI to display "undefined Data", this change will simply show it as "Data"

### DIFF
--- a/client/src/pages/Tree/TreeLinks.js
+++ b/client/src/pages/Tree/TreeLinks.js
@@ -10,11 +10,11 @@ export default function TreeLinks({ currentTreeData }) {
   const links = [
     {
       url: currentTreeData.download,
-      label: `${currentTreeData.city} Data`,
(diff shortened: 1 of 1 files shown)


jazhen:
@zoobot - Can you review this one?

zoobot (commented):
If it doesn't have currentTreeData.city, it could have currentTreeData.sourceId, which is sort of the same thing, so I'd maybe do something like this:

const { download, info, city, sourceId } = currentTreeData;

label: `${city || sourceId || ''} Data`,

Or tackle it upstream if MapLayout.js line 62 is not working correctly:
https://github.com/waterthetrees/wtt_front/blob/main/client/src/pages/Map/MapLayout.js#L62

anup-a:
@zoobot I've updated as per your suggested approach

zoobot (approved):
Thanks for your contribution!

PR #347, opened 2022-10-27
patch: fix scrolling issue of privacy policy page

chinmaykumbhare:


chinmaykumbhare:
hi @jazhen can you please review?

jazhen (commented):
Hi, we wanted the whole page to be scrollable rather than just the one section. There is a root issue that needs to be addressed for the entire app.

Yours:
https://user-images.githubusercontent.com/6326660/198374811-08159ad7-3254-49a6-8163-b711a4c48306.mov

Desired:
https://user-images.githubusercontent.com/6326660/198374755-cffb0d76-ca06-40f2-b1f8-80f5bdf1d11c.mov

jazhen:
@vernfongchao - I think you mentioned an issue with removing the `overflow: hidden` line in [App.css#L55](https://github.com/waterthetrees/wtt_front/blob/main/client/src/components/App/App.css#L55). Can you explain that again?

vernfongchao:
@jazhen yeah, so I didnt mention the issue with removing overflow:hidden but something I noticed when tackling not being able to scroll issue. Rose was the one that mentioned that it was added due to the map. I'm not exactly sure what conflicts would be caused by removing it.

chinmaykumbhare:
@jazhen removing / modifying a css property from parent component is always a bad idea. Instead, what I would suggest is we make the component scrollable

jazhen:
> @vernfongchao - I think you mentioned an issue with removing the `overflow: hidden` line in [App.css#L55](https://github.com/waterthetrees/wtt_front/blob/main/client/src/components/App/App.css#L55). Can you explain that again?

  > @jazhen yeah, so I didnt mention the issue with removing overflow:hidden but something I noticed when tackling not being able to scroll issue. Rose was the one that mentioned that it was added due to the map. I'm not exactly sure what conflicts would be caused by removing it.

@zoobot - Can you explain why we need the overflow hidden line? I didn't see any obvious issues when removing that line.

jazhen:
> @jazhen removing / modifying a css property from parent component is always a bad idea. Instead, what I would suggest is we make the component scrollable

I understand the hesitation since there may be unintended consequences. But this is a root issue that will affect every new page that we make. If only the map needs the `overflow: hidden` css then it should be moved from the parent css to the map component specifically.

chinmaykumbhare:
@jazhen can we set it to overflow : auto in App.css and fix the issues that occur as a result?

chinmaykumbhare:
@jazhen I can see no visible issues with these new changes. Let me know if I am missing something.

chinmaykumbhare:
@zoobot can you send across a screenshot? For my screen, the Map is not scrollable with respect to my recent changes.

chinmaykumbhare:
@zoobot I get your concern, instead of using styled components everywhere we would set the CSS property of Map page to initial one.

@jazhen It would be really helpful if you can approve

jazhen:
> Will making the privacy policy a styled component to fix this issue?

@zoobot - It won't because `overflow: hidden` is set as a global attribute on the `body`.

client/src/pages/Map/Map.css
```diff
@@ -1,3 +1,7 @@
+body {
+  overflow: hidden !important;
+}
+
```
jazhen, line 4: I don't love modifying the `body` not in the global styles file.

client/src/components/App/App.css
```diff
@@ -52,7 +52,7 @@ body {
   width: 100vw;
   height: 100vh;
   background-color: #fff;
-  overflow: hidden;
+  overflow: auto;
```
jazhen, line 55: I think the only change we need for the PR is to remove this `overflow` line. Will add a video of what I think this is trying to do.

vernfongchao:
from what I can tell, overflow is set to hidden because of the hamburger menu that comes out from clicking the navbar button. When its overflow hidden is removed and when the menu is open it seems to add width and height that extends the page causing scroll bars to appear.

jazhen:
Essentially, when `margin: 0` is off and `overflow: hidden` is off, you can scroll the page because there is extra margin causing overflow. 

When `margin: 0` is off and `overflow: hidden` is on, we also can't scroll the page, but there is extra margin being hidden now.

So we want to keep `margin: 0`, but we don't need `overflow: hidden`.

cc: @zoobot - That's what I've found. Not sure if you know of a different reason to have `overflow: hidden`.

https://user-images.githubusercontent.com/6326660/198900111-0e11d5bb-d817-497a-b6d1-c74aea70b552.mov

vernfongchao:
I think the overflow hidden was originally used to hide the horizontal scrollbar was caused due to setting body's width to 100vw. I remember reading something about why you should used % instead on the body's width instead. I noticed what behavior it was that it was adding horizontal scrollbars to the body because 100VW includes the width of scrollbar. The map page isn't affected by it but the others are. When changing it to percentage the scrollbar issue goes away


image
image

image
image

jazhen:
@vernfongchao - Nice find, I've had that problem before 🙁, but have mostly avoided viewport since so didn't even think of that. For me `height: 100vh` is showing the vertical scrollbar when there should be one, but there's no horizontal scrollbar whether it's `width: 100vw` or `width: 100%`.

chinmaykumbhare:
@jazhen @vernfongchao can you guys help me out in understanding the exact changes that I need to make?

vernfongchao:
@chinmaykumbhare I believe we should put this pr on draft until we are able to have a more in depth discussion about this on Wednesday's meeting, as of now this problem is apparent on all of the pages. I believe a solid fix on whatever is causing the  underlying issue is better for us moving forward/scale than giving a temporary fix.

zoobot (changes requested):
@chinmaykumbhare if we remove overflow: hidden from App.css, the issue is fixed. 
Can you make that change?
Thanks @chinmaykumbhare @vernfongchao @jazhen !!!

vernfongchao:
The fix I suggest was to remove the overflow hidden from root div and change the width and height from vw and vh to percentages. That seems to have fix the issues of seeing unnecessary scrollbars that's not needed @zoobot

Issue #340, opened 2022-10-26
Fix ESLint issues (Errors) - jsx-a11y/alt-text


jazhen:
> Running `npm run lint` in the command line shows various ESLint issues that we should fix. The issues are split into warnings (yellow-coded) and errors (red-coded). Within warnings and errors, the issues are split into categories by ESLint rule name.
> 
> Example: ![image](https://user-images.githubusercontent.com/6326660/197923380-4c777e9b-8a8b-461b-8839-cde324be0a72.png)

Fix all errors under the category `jsx-a11y/alt-text`. _Please make sure to regression test so that we know nothing has broken through these changes._

List of issues (May not be exhaustive so double check):

- [ ] `wtt_front/client/src/pages/Tree/TreeMaintenanceDialog.js`
  - [ ]   56:13  error  img elements must have an alt prop, either with meaningful text, or an empty string for decorative images  jsx-a11y/alt-text

DamyanBG:
Please, assign me the issue.

jazhen:
> Please, assign me the issue.

Assigned. Thank you!

jazhen:
Closed by #351

Issue #342, opened 2022-10-26
Fix ESLint issues (Warnings) - no-console


jazhen:
> Running `npm run lint` in the command line shows various ESLint issues that we should fix. The issues are split into warnings (yellow-coded) and errors (red-coded). Within warnings and errors, the issues are split into categories by ESLint rule name.
> 
> Example: ![image](https://user-images.githubusercontent.com/6326660/197923380-4c777e9b-8a8b-461b-8839-cde324be0a72.png)

Fix all warnings under the category `no-console`. _Please make sure to regression test so that we know nothing has broken through these changes._

List of issues (May not be exhaustive so double check):

- [ ] `wtt_front/client/src/api/queries.js`
  - [ ]    68:9   warning  Unexpected console statement                                                                                                                                                                                                                        no-console
- [ ] `wtt_front/client/src/components/ErrorBoundary/ErrorBoundary.js`
  - [ ]   7:3  warning  Unexpected console statement  no-console
- [ ] `wtt_front/client/src/pages/NewTree/NewTree.js`
  - [ ]   56:5   warning  Unexpected console statement                                                                                                                                                                no-console
- [ ] `wtt_front/client/src/pages/Tree/TreeEdit.js`
  - [ ]   98:40  warning  Unexpected console statement  no-console
- [ ] `wtt_front/data/build.js`
  - [ ]    38:3  warning  Unexpected console statement  no-console
  - [ ]   227:3  warning  Unexpected console statement  no-console

jazhen:
Getting similar errors to the following after importing `logger` from `wtt_front/server/logger.js`. I believe this because we are trying to use node packages which not meant to be used in the browser.

```
ERROR in ./node_modules/winston/dist/winston/transports/http.js 39:12-28
Module not found: Error: Can't resolve 'https' in '/Users/jzhen/Dev/waterthetrees/wtt_front/node_modules/winston/dist/winston/transports'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "https": require.resolve("https-browserify") }'
        - install 'https-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "https": false }
 @ ./node_modules/winston/dist/winston/transports/index.js 41:11-28
[...]

PR #351, opened 2022-10-31
delete console log 

DamyanBG:
I fixed ESLint issue, which is described in #340

### DIFF
--- a/client/src/pages/Tree/TreeEdit.js
+++ b/client/src/pages/Tree/TreeEdit.js
@@ -95,12 +95,15 @@ export default function TreeEdit({
     onConfirm(formData);
   };
 
+  const handleError = (errorLog, e) => console.error('errors, e', errorLog, e);
+
   return (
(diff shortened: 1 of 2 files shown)


client/src/pages/Tree/TreeEdit.js
```diff
@@ -95,15 +95,12 @@ export default function TreeEdit({
...
 
-  const handleError = (errorLog, e) => console.error('errors, e', errorLog, e);
-
   return (
     <FormScrollableDialog
       title="Edit Tree"
       open={open}
       onConfirm={handleConfirm}
       onCancel={onCancel}
-      onError={handleError}
```
jazhen, line 106: I don't think these changes should be part of this PR. These would be part of #342.

jazhen (changes requested):

jazhen (approved):
Thanks for making the requested changes. Looks good!

PR #362, opened 2022-11-03
fixed issue 361 useForm

### LINKED ISSUE #360: 21:25  error  React Hook "useAuth0" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?  react-hooks/rules-of-hooks

vernfongchao:
#361 
Before:
image

Fix
image

The lint issue was caused by a condition returning early before a hook was called.
What I did was move that condition to the bottom of the hook calls from useAuth0 and useForm. 

Then to make sure there wasn't any issue with key into newTreeState.coords for the lng and lat key for defaultValues. I used a ternary to see if it exists I can key into lng and lat, but if it doesnt it sets it to an empty string. 

This way if it doesnt exists it wont error out and will be set to an empty string. but will still return null after when it doesn't exist. It might not be the most efficient since you're still end up creating defaultValues, but this will remove that error.

I have tested it and it doesn't seem to be causing any errors when placing down a marker and clicking on it and is working the same way as before. I havent actually tried to plant since I don't want to mess with the map.

@zoobot I would love it if you could look at this, since it pertains to the map.
@Tijani-zainab This would also fix #360 as well

### DIFF
--- a/client/src/pages/NewTree/NewTree.js
+++ b/client/src/pages/NewTree/NewTree.js
@@ -18,7 +18,7 @@ export default function NewTree({ TreeDetailsContainer, drawerWidth }) {
   const { nickname, email, name } = user;
 
   /* 
-    tenery used to avoid any errors caused by useForm hook
+    ternary used to avoid any errors caused by useForm hook
     when moving early return statements after hook calls for linting
   */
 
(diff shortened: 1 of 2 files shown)


Tijani-zainab:
Nice Work ! @vernfongchao 
referencing this issue #360 , Thank You !

Tijani-zainab:
Oh, this fixes #360 as well 👍🏽

zoobot (approved):
Nice catch! Please leave a TODO/REASON for the ternary and then feel free to merge. Thanks!  @vernfongchao

vernfongchao:
Added comment for ternary
image

Issue #192, opened 2022-01-10
default center of map to user's browser data?


zoobot:

fwextensions:
I'd say no, if it's going to ask for the user's location as soon as they hit the website.  I'd bounce at that point.  But if/when the user has given the location permission, then it probably makes sense to default them to their current location, if there's no previous location available (like a hash in the URL).

fwextensions:
This is one of the services that does IP to geolocation lookup: https://ipstack.com/  100/month for free, 50K/month for $10/month.

mwpark2014:
@fwextensions curious on if you still have opinions on this? IMO, asking users for location permissions is a better interaction than requiring the user to zoom in from a zoomed out global view. The Geolocation API would probably be a good low-cost solution that we can test and we can try ipstack.com if we need better IP tracking.

Other options:
Try to guess the user's preferred viewport by storing user's most recent viewports or tree planting data from profile.

zoobot:
@mwpark2014 
Geolocation API's a good idea for first round on this.

If they already have location services allowed AND if there's no pos/zoom hash in the url, 
let's center on general globe location. 
Don't ask on initial load. If it's not allowed already, do nothing.

mwpark2014:
@zoobot 
Thanks for the response! Okay, as I understand it:

v0:
If they already have location services allowed AND if there's no pos/zoom hash in the url, use Geolocation API to center map

v1:
As a fallback, let's use ipstack to get general location to center map

Additional questions: 
Is there currently only one way to ack for location services permission, the button here: 
![image](https://user-images.githubusercontent.com/10445556/209234156-7d36fe5a-af2a-48ea-b492-047d3cd7d070.png)?

Is there a reason why we don't already automatically increase the zoom level once we manually center on the user's location? I think it's cool to zoom out to the globe, but wonder if it would save a few clicks/scrolls to zoom in a couple levels.

zoobot:
V0 and V1 sound good.
Yes that is the way we are currently asking for location services permission.

The zoom is a bit jarring and messed up currently, imo. 
If you want to make the whole zoom thing smoother, we have a zoom smoothing issue opened.
https://github.com/waterthetrees/wtt_front/issues/276

I can see being zoomed in more would be nicer if we have location, especially on mobile.

mwpark2014:
Closing since this issue is being replaced by:
1) https://github.com/waterthetrees/wtt_front/issues/392 for first time users
[...]

PR #378, opened 2022-12-23
Use Geolocation API to center map by current location on load in certain cases

mwpark2014:
See https://github.com/waterthetrees/wtt_front/issues/192

This is 
> v0: If they already have location services allowed AND if there's no pos/zoom hash in the url, use Geolocation API to center map

Unfortunately, it appears that checking if a user has already given location permissions is only possible through an async promise. Therefore, the options are either making the map load wait until we can pass in user's geo coords or recentering the map after the mapBox object has been created. 
In my approach, I try to kick off the permissions query asap. Inside a `useEffect` hook, we recenter the map as soon as the permissions query promise resolves and there is a `map` object in state. This is not guaranteed to run before `mapboxMap` kicks off its `load` event.

Assumptions:
* This app only supports modern browsers due to support for the Permissions API:  https://caniuse.com/permissions-api
* This code is only ran client-side so we don't run into any node shenanigans

Manually Tested:
* User does not get prompted for location permissions
* If there is a hash coordinate already present, map will not recenter
* If user does have permissions and there is no existing hash coord, map will load up centered at user's location

### DIFF
--- a/client/src/pages/Map/Map.js
+++ b/client/src/pages/Map/Map.js
@@ -35,9 +35,11 @@ const unsupportedError = (
   </Box>
 );
 
-// If browser supports geolocation, let's check if we have geo permissions and update lat/long if so
+// If browser supports geolocation and there is no hash param on page load,
+// let's check if we have geo permissions and use that to recenter map if so
 const geoPermissionsPromise =
-  'geolocation' in navigator
+  'geolocation' in navigator &&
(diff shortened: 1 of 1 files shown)


mwpark2014:
Closing since this issue is being replaced by:
1) https://github.com/waterthetrees/wtt_front/issues/392 for first time users
2) https://github.com/waterthetrees/wtt_front/issues/447 for repeat users

PR #385, opened 2022-12-26
Fix bug: When viewing tree details, geolocate feature breaks

### LINKED ISSUE #384: Bug: When viewing tree details, geolocate feature breaks. Every subsequent activation recenters upon previously selected tree coordinates
Repro steps:
* Activate geolocation using the geolocate icon button on the right
* Select a tree
* Unselect the tree by clicking on the map
* Deactivate and reactivate geolocation. The map recenters on previously selected tree instead of user location

TODO: Show a GIF of issue

mwpark2014:
Fixes #384

### DIFF
--- a/client/src/pages/Map/Map.js
+++ b/client/src/pages/Map/Map.js
@@ -194,7 +194,10 @@ export default function Map({
             hashParams.delete('id');
           }
 
-          window.location.hash = decodeURIComponent(hashParams.toString());
+          const newUrl = `${window.location.origin}#${decodeURIComponent(
+            hashParams.toString(),
+          )}`;
+          window.history.replaceState({}, '', newUrl);
         });
 
         // Unlike the click handler above, we want to get mousemove/leave events only for features
--- a/client/src/pages/Map/Map.js
+++ b/client/src/pages/Map/Map.js
@@ -190,6 +190,7 @@ export default function Map({
           } else {
             // This click was on a blank part of the map, so clear the selection.
             setCurrentTreeId(null);
+            setCurrentTreeDataVector(null);
             hashParams.delete('id');
           }
 
--- a/client/src/pages/Map/Map.js
+++ b/client/src/pages/Map/Map.js
@@ -1,4 +1,5 @@
 import React, { useState, useEffect, useRef } from 'react';
+import { unstable_batchedUpdates } from 'react-dom';
 import mapboxgl from 'mapbox-gl';
 import { Alert, Box, Typography } from '@mui/material';
 import { useQueryClient } from 'react-query';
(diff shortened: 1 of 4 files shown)


mwpark2014:
Before:
RecenterStuck-comp

After:
GeolocateVsTreeCenter-comp

Latitude/longitude is untouched in the url hash.

I would expect it to cause flyto only in reaction to a user activating geolocate or clicking on a tree. It should not flyto when deactivating geolocate - look like that is not the current updated functionality so let me iterate on this a bit @zoobot

mwpark2014:
@zoobot  Fixing this bug turned into quite the rabbit hole! As a result, this PR has grown in complexity and is something I'd like to walk someone through in the next meeting.

Summary of changes
* When viewing tree details, geolocate feature recenters over the previously viewed tree
  * Issue: Turns out, this is due to the useEffect in MapLayout (https://github.com/waterthetrees/wtt_front/blob/c7007e579d1ad3358ecc09fd13d7b229b3333ce6/client/src/pages/Map/MapLayout.js#L79-L81 ) _always_ firing upon rerender. This is because `hashParams` is always a new `URLSearchParams` and always the triggers the `useEffect` even if all the deps are unchanged. As a result, the map overrides the `flyTo` to the user's current geolocation with a second `flyTo` to the currently selected tree location
  * Solution: remove `hashParams` from the useEffect deps
  * This exposed another issue: When clicking on a tree, the map doesn't recenter to the tree at all. This turns out to be due to the setting of hashes (https://github.com/waterthetrees/wtt_front/blob/64a4e567616f0c886e2a5f954f8ffc5287fc3d2f/client/src/pages/Map/Map.js#L196). Whenever a hash is set, it will interrupt a `flyTo` operation in motion. This wasn't noticed before since MapLayout is rerendered many times and its `useEffect`s were rerendered many times due to the above `hashParams` issue. Essentially, many `map.flyTo` functions were called, interrupted, and then called again, so it _seemed_ like it was working. 
  * Solution: Instead of setting `window.location.hash`, call `window.replaceState` to update the URL without causing any mapbox side effects. **This has significant changes on smoothness of mapbox interactivity including zooms no longer being cancelled and handleTransitionEnd no longer being necessary (according to initial testing)**
* Now, when a currently selected tree is deselected (by clicking on the map), the map recenters to the previously selected tree location
  * Solution: Make sure to set `currentTreeDataVector` to null in addition to `currentTreeId` when deselecting
  * This exposed another issue: React 17 doesn't batch state updates from native event listeners, so `setCurrentTreeId` and `setCurrentTreeDataVector` will cause MapLayout to rerender separately causing a phantom `flyTo` to the old coordinates
  * Solution: Manually batch these state changes together. This was the quickest fix, but better solutions exist

Open to suggestions, especially on test cases I probably haven't considered! I have some ideas for refactoring as well, but I have decided to hold off on any sort of refactoring until further discussion or a separate PR entirely

mwpark2014:
Closing this PR, since I'm breaking this down into smaller pieces / changing the approach

Issue #321, opened 2022-10-18
Implement Location based weather


ri0nardo:

DamyanBG:
This issue seems interesting, I want to work on it.

jazhen:
@DamyanBG - Sure, I'll assign you. I think the task to create a new weather component as circled in red. Our user profile page does not currently match the designs, so right now just create a reusable component that we can insert when everything's ready. 

![image](https://user-images.githubusercontent.com/6326660/199763357-61801b91-57a4-49bc-8343-75ac4899d034.png)

DamyanBG:
I am from the team and i assigned it to myself.

ri0nardo:
@DamyanBG any progress on the location based weather? Currently thinking of all types of weather designs.

DamyanBG:
@ri0nardo is the user profile page updated ?

DamyanBG:
Here is my mock up. The section is almost ready. I have few questions:

- I am using OpenWeatherAPI and we have to store API_KEY, describe me the process behind that
- Do you want the units to be metric or another kind?

PR #387, opened 2023-01-03
Weather Section

DamyanBG:
I implemented #321

### DIFF
--- a/client/src/pages/Userprofile/UserWeatherSection.js
+++ b/client/src/pages/Userprofile/UserWeatherSection.js
@@ -1,0 +1,142 @@
+import React, { useEffect, useState } from 'react';
+import { useGeolocation } from '../UserLocation/useGeolocation';
+
+export default function UserWeatherSection() {
+  const date = new Date();
+  const months = [
+    'January',
+    'February',
+    'March',
+    'April',
+    'May',
+    'June',
+    'July',
+    'August',
+    'September',
+    'October',
+    'November',
+    'December',
+  ];
+  const weekdays = [
+    'Sunday',
+    'Monday',
+    'Tuesday',
+    'Wednesday',
+    'Thursday',
+    'Friday',
+    'Saturday',
+  ];
+
+  const month = months[date.getMonth()];
+  const weekday = weekdays[date.getDay()];
+  const day = date.getDay();
+
+  const { data } = useGeolocation();
+  const API_KEY = 'e28a2bf635aa01de83b389bce17192cc';
+
+  const [temperature, setTemperature] = useState(null);
+  const [weather, setWeather] = useState(null);
+  const [city, setCity] = useState(null);
+  const [country, setCountry] = useState(null);
+  const [windSpeed, setWindSpeed] = useState(null);
+  const [compassDirection, setCompassDirection] = useState('');
+
+  const fetchWeather = async () => {
+    if (!data || temperature) return;
+    const response = await fetch(
+      `http://api.openweathermap.org/data/2.5/weather?units=metric&lat=${data?.coords?.latitude}&lon=${data?.coords?.longitude}&appid=${API_KEY}`,
+    );
+    const json = await response.json();
+    console.log(json);
+    setTemperature(json.main.temp);
+    setWeather(json.weather[0].main);
+    setCity(json.name);
+    setCountry(json.sys.country);
+    setWindSpeed(json.wind.speed);
+    const points = Math.round(json.wind.deg / 22.5);
+    const directions = [
+      'N',
+      'NNE',
+      'NE',
+      'ENE',
+      'E',
+      'ESE',
+      'SE',
+      'SSE',
+      'S',
+      'SSW',
+      'SW',
+      'WSW',
+      'W',
+      'WNW',
+      'NW',
+      'NNW',
+    ];
+    const index = points % 16;
+
+    setCompassDirection(directions[index]);
+  };
+
+  useEffect(() => {
+    fetchWeather();
+  }, [data]);
+
+  console.log(data);
+
+  return (
+    <section
+      style={{
+        margin: '48px',
+        width: '300px',
+        height: '200px',
+        borderRadius: '6px',
+        backgroundColor: 'rgb(127, 210, 244)',
+        color: 'white',
+        padding: '20px',
+        display: 'flex',
+        flexDirection: 'column',
+      }}
+    >
+      <p>{weather}</p>
+      <div style={{ display: 'flex' }}>
+        <article
+          style={{
+            fontSize: '32px',
+            borderRight: '1px solid white',
+            paddingRight: '16px',
+            marginRight: '16px',
+          }}
+        >
+          {temperature}&deg;
+        </article>
+        <article style={{ display: 'flex', alignItems: 'center' }}>
+          {weekday}, {day} {month}
+          <br />
+          {city}, {country}
+        </article>
+      </div>
+      <div
+        style={{
+          display: 'flex',
+          justifyContent: 'space-between',
+          marginTop: 'auto',
+          marginBottom: '0',
+          alignSelf: 'bottom',
+        }}
+      >
+        <article>
+          <p>Wind Speed</p>
+          <p style={{ fontSize: '20px' }}>{windSpeed}km/h</p>
+        </article>
+        <article>
+          <p>Wind Direction</p>
+          <p style={{ fontSize: '20px' }}>{compassDirection}</p>
+        </article>
+        <article>
+          <p>UVI</p>
+          <p style={{ fontSize: '20px' }}>10</p>
+        </article>
+      </div>
+    </section>
+  );
+}
--- a/client/src/pages/Userprofile/UserWeatherSection.js
+++ b/client/src/pages/Userprofile/UserWeatherSection.js
@@ -1,86 +1,64 @@
 import React, { useEffect, useState } from 'react';
 import { useGeolocation } from '../UserLocation/useGeolocation';
 
+const directions = [
+  'N',
+  'NNE',
+  'NE',
+  'ENE',
+  'E',
+  'ESE',
+  'SE',
+  'SSE',
+  'S',
+  'SSW',
+  'SW',
+  'WSW',
+  'W',
+  'WNW',
+  'NW',
+  'NNW',
+];
+
 export default function UserWeatherSection() {
   const date = new Date();
-  const months = [
-    'January',
-    'February',
-    'March',
-    'April',
-    'May',
-    'June',
-    'July',
-    'August',
-    'September',
-    'October',
-    'November',
-    'December',
-  ];
-  const weekdays = [
-    'Sunday',
-    'Monday',
-    'Tuesday',
-    'Wednesday',
-    'Thursday',
-    'Friday',
-    'Saturday',
-  ];
+  const month = date.toLocaleString('default', { month: 'long' });
+  const weekday = date.toLocaleString('default', { weekday: 'long' });
+  const day = date.getDate();
 
-  const month = months[date.getMonth()];
-  const weekday = weekdays[date.getDay()];
-  const day = date.getDay();
+  const { data } = useGeolocation({ watching: false });
 
-  const { data } = useGeolocation();
+  const [weather, setWeatherAPI] = useState({})
 
-  const [temperature, setTemperature] = useState(null);
-  const [weather, setWeather] = useState(null);
-  const [city, setCity] = useState(null);
-  const [country, setCountry] = useState(null);
-  const [windSpeed, setWindSpeed] = useState(null);
-  const [compassDirection, setCompassDirection] = useState('');
+  const findDirection = (degrees) => {
+    const points = Math.round(degrees / 22.5)
+    const index = points % 16;
+    const direction = directions[index]
+    return direction
+  }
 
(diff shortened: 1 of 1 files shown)


DamyanBG:
for review

DamyanBG:
Guys, help me to refactor the code. Every team has it his coding way, so give me suggestions how to split the code.

I need the information about storing the API_KEYS. For weather based information I have to use 3rd party service and they have access keys.

Open Weather do not have UVI information, so do I have to change the API provider or to hide UVI?

client/src/pages/Userprofile/UserWeatherSection.js
```diff
@@ -0,0 +1,141 @@
...
+    'Sunday',
+    'Monday',
+    'Tuesday',
+    'Wednesday',
+    'Thursday',
+    'Friday',
+    'Saturday',
+  ];
+
+  const month = months[date.getMonth()];
```
mwpark2014, line 30: Just calling it out that you can use internationalization API: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString

`const month = date.toLocaleString('default', { month: 'long' });`
`const weekday = (new Date()).toLocaleString('default', { weekday: 'long' });`

A benefit of doing it your way though is that it'll be compatible on all browser versions!

client/src/pages/Userprofile/UserWeatherSection.js
```diff
@@ -0,0 +1,141 @@
...
+  const weekday = weekdays[date.getDay()];
+  const day = date.getDay();
+
+  const { data } = useGeolocation();
```
mwpark2014, line 34: There are a couple concerns about using `useGeolocation` here. 

https://github.com/waterthetrees/wtt_front/blob/64a4e567616f0c886e2a5f954f8ffc5287fc3d2f/client/src/pages/UserLocation/useGeolocation.js#L4-L5
1) The hook will ask for user's permission upon render of the component, since `useGeolocation` enables geolocation by default. This might be disruptive depending on how/where you add in `UserWeatherSection` (for example, if `UserWeatherSection` is called upon first load of the map)
2) By default, watching is set to true, which means that updates will continuously be sent to `data` here. This is a problem because `fetchWeather` is called whenever `data` is changed and will cause a ton of calls to `api.openweathermap`. 
For this, I suggest sending `useGeolocation({ watching: false })`

client/src/pages/Userprofile/UserWeatherSection.js
```diff
@@ -0,0 +1,141 @@
...
+    );
+    const json = await response.json();
+    console.log(json);
+    setTemperature(json.main.temp);
+    setWeather(json.weather[0].main);
+    setCity(json.name);
+    setCountry(json.sys.country);
+    setWindSpeed(json.wind.speed);
+    const points = Math.round(json.wind.deg / 22.5);
+    const directions = [
```
mwpark2014, line 56: Prefer the constants `directions`, `months`, `weekdays` to be at the top of the file outside of `UserWeatherSection` for readability

mwpark2014 (changes requested):

mwpark2014:
I'll leave it to @zoobot to answer the question about where to store API keys

client/src/pages/Userprofile/UserWeatherSection.js
zoobot: To avoid re-renders, you may want to just `setWeatherAPIData(json)` for one single state change with all the columns you need.
`const [weather, setWeatherAPIData] = useState(null);`

To validate the response and check that those columns actually exist, you could use json?.main?.temp or destructuring so it doesn't crash when say json.wind.deg is undefined. Also you could map through the weather object.

It would be nice to cache the data so it doesn't call the weatherapi with every load.
This has a nice idea for caching, in our case we could cache by city with a date field or something to cut down on api calls: 
https://medium.com/@uurylmaz_62733/how-to-cache-data-in-react-local-storage-dd2992e053d8

client/src/pages/Userprofile/UserWeatherSection.js
DamyanBG: resolved

client/src/pages/Userprofile/UserWeatherSection.js
DamyanBG: resolved

client/src/pages/Userprofile/UserWeatherSection.js
```diff
@@ -0,0 +1,141 @@
...
+  const [windSpeed, setWindSpeed] = useState(null);
+  const [compassDirection, setCompassDirection] = useState('');
+
+  const fetchWeather = async () => {
+    if (!data || temperature) return;
+    const response = await fetch(
+      `http://api.openweathermap.org/data/2.5/weather?units=metric&lat=${data?.coords?.latitude}&lon=${data?.coords?.longitude}&appid=${API_KEY}`,
+    );
+    const json = await response.json();
+    console.log(json);
```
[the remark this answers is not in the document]
DamyanBG: resolved

client/src/pages/Userprofile/UserWeatherSection.js
DamyanBG: I made the required changes.

DamyanBG:
My PR is ready for review.

ri0nardo:
@DamyanBG has anyone reviewed this?

DamyanBG:
> @DamyanBG has anyone reviewed this?

No.

mwpark2014:
Sorry about that. I'll take a look tomorrow

client/src/pages/Userprofile/UserWeatherSection.js
```diff
@@ -0,0 +1,147 @@
...
+export default function UserWeatherSection() {
+  const currentDate = new Date();
+  const month = currentDate.toLocaleString('default', { month: 'long' });
+  const weekday = currentDate.toLocaleString('default', { weekday: 'long' });
+  const day = currentDate.getDate();
+
+  const { data } = useGeolocation({ watching: false });
+
+  // Pleae, add your API_KEY
+  const API_KEY = 'write_the_key_here';
```
mwpark2014, line 72: Is this meant to be updated?

client/src/pages/Userprofile/UserWeatherSection.js
```diff
@@ -0,0 +1,147 @@
...
+    return direction;
+  };
+
+  const fetchWeather = async () => {
+    const weatherStorage = JSON.parse(localStorage.getItem(WEATHER_CACHE));
+    if (weatherStorage && weatherStorage.expiry > currentDate) {
+      setWeatherAPI(weatherStorage.weather);
+      return;
+    }
+    const response = await fetch(
```
mwpark2014, line 89: What will happen if the `fetch` ends up returning an error? Probably need to handle any possible exceptions with a try catch

client/src/pages/Userprofile/UserWeatherSection.js
```diff
@@ -0,0 +1,147 @@
...
+  const day = currentDate.getDate();
+
+  const { data } = useGeolocation({ watching: false });
+
+  // Pleae, add your API_KEY
+  const API_KEY = 'write_the_key_here';
+
+  const [weather, setWeatherAPI] = useState({});
+
+  const findDirection = (degrees) => {
```
mwpark2014, line 76: This can be extracted outside of the react component as well

client/src/pages/Userprofile/UserWeatherSection.js
```diff
@@ -0,0 +1,147 @@
...
+      country: json?.sys?.country,
+      windSpeed: json?.wind?.speed,
+      compassDirection: compassDirection,
+    };
+
+    setWeatherAPI(newWeather);
+    localStorage.setItem(
+      WEATHER_CACHE,
+      JSON.stringify({
+        expiry: currentDate.getTime() + ONE_HOUR,
```
mwpark2014, line 109: I wonder if we can go even more aggressive with this caching. How did you pick one hour? This is merely a discussion point and I'm not saying it's a bad value to start with

mwpark2014:
Left some comments. Where will `UserWeatherSection` show up on the map?

mwpark2014 (changes requested):

client/src/pages/Userprofile/UserWeatherSection.js
```diff
@@ -0,0 +1,147 @@
...
+
+  const findDirection = (degrees) => {
+    const points = Math.round(degrees / 22.5);
+    const index = points % 16;
+    const direction = directions[index];
+    return direction;
+  };
+
+  const fetchWeather = async () => {
+    const weatherStorage = JSON.parse(localStorage.getItem(WEATHER_CACHE));
```
mwpark2014, line 84: I like it! 👍

client/src/pages/Userprofile/UserWeatherSection.js
```diff
@@ -0,0 +1,147 @@
...
+  'NNW',
+];
+
+export default function UserWeatherSection() {
+  const currentDate = new Date();
+  const month = currentDate.toLocaleString('default', { month: 'long' });
+  const weekday = currentDate.toLocaleString('default', { weekday: 'long' });
+  const day = currentDate.getDate();
+
+  const { data } = useGeolocation({ watching: false });
```
mwpark2014, line 69: `data` is a little non-descript. Could we name this variable more specifically so developers don't confuse this value with the current map position's latitude / longitude?

client/src/pages/Userprofile/UserWeatherSection.js
DamyanBG: Yes, as I wrote I am using my API_KEY, I think it is better the project owners to create API_KEY for the project and use it.

client/src/pages/Userprofile/UserWeatherSection.js
DamyanBG: OK

client/src/pages/Userprofile/UserWeatherSection.js
DamyanBG: Why

client/src/pages/Userprofile/UserWeatherSection.js
DamyanBG: I dont remember the exact decision why I took one hour, but what do you mean about more aggressive?

client/src/pages/Userprofile/UserWeatherSection.js
DamyanBG: I can't understand what do you mean?

client/src/pages/Userprofile/UserWeatherSection.js
mwpark2014: Yup, I see that zoobot responded above
> Does open weather have an allow list of urls for the key? We usually add only our dev.waterthetrees.com and waterthetrees.com to the allow list once we are done testing. You can add the key here once it's locked down to only our urls: https://github.com/waterthetrees/wtt_front/blob/main/client/src/util/config.js

Did you see this response? It looks like step 1: configure the account tied to the API key to only allow a whitelist of URLs to use it and step 2: commit the key to `config.js`. Can you try this?

If you can't create a whitelist, then we need to store the API key within an environment variable.

client/src/pages/Userprofile/UserWeatherSection.js
mwpark2014: `findDirection` is a constant function that doesn't need to be reinstantiated every time this React component is rerendered. There doesn't seem to be any dependencies inside `findDirection` that depends on React states or props

client/src/pages/Userprofile/UserWeatherSection.js
mwpark2014: What I mean by more aggressive is potentially doing 2 hours, 6 hours, or 24 hours. 1 hour is okay for now.

client/src/pages/Userprofile/UserWeatherSection.js
mwpark2014: This `data` variable has a latitude and longitude. When using this variable elsewhere in the file, it is quite easy to misunderstand what `data` means. It represents some sort of location... could be the map's current position or even the tree's position.

```suggestion
  const { data: userLocation } = useGeolocation({ watching: false });
```

DamyanBG:
I don't want to contribute to this project anymore.

DamyanBG:
Hello,

I can see nobody is still working on that task. 

I can finish it, but first - I need more information how you store the env keys!

If that was my project, I would not store the key for weather service in the front end. I would not make any requests to the weather  API service from the front end. Instead, I would make an api route to the back end, which is called from the front end for the weather information.

At the back end I would implement an adapter pattern for the weather information service. That is because in future I can change the provider for the weather api and it would be easier if I have an adapter pattern.

Then I would cache a little bit the information about the weather at the back end. If I have requests for the same place, I would return the cached information. That will limit the requests to the weather service API, since there can be paid or limited plan.

After that I would cache the weather information on the front end too. No need to make many requests to the back end for the same information.

So my questions to you are:
 
1. Who can provide me information about the environments variables?
2. Do you like my idea about the creating the logic for requesting weather information from the back end?
3. Where you would want me to cache the weather information in the FE? In the localStorage, sessionStoraga, IndexedDB??

@zoobot @ri0nardo @mwpark2014

Issue #366, opened 2022-11-28
Implement new header design for static screens


ri0nardo:

rabbanirizwan:
@ri0nardo kindly assign me this and also share the Figma link

ri0nardo:
@rabbanirizwan assigning you to this ticket. Please read the notes before beginning. Figma is linked with the ticket

zoobot:
@rabbanirizwan Any update on this? Are you still working on this or should we re-assign it? Thanks!!

ri0nardo:
@vernfongchao no response from the current assignee

vernfongchao:
@ri0nardo 
![image](https://user-images.githubusercontent.com/91238232/212438122-2771b999-b1ed-4ba8-be6f-ce8081d1c5b6.png)
![image](https://user-images.githubusercontent.com/91238232/212438153-69ee3336-3608-43bc-bb2e-871a85b68ace.png)

hovered
![image](https://user-images.githubusercontent.com/91238232/212438185-df40fcc3-7c0d-4f54-ac61-85ad32895ca3.png)
Logged In (full screen)
![image](https://user-images.githubusercontent.com/91238232/212438268-65a531ce-f73f-4f10-86fe-96b07ff1323f.png)
Logged In (mobile)
![image](https://user-images.githubusercontent.com/91238232/212438309-5e0fb72e-e63d-43ac-a703-89d389258708.png)
Logged out (mobile)
![image](https://user-images.githubusercontent.com/91238232/212438352-8fd9b75d-cfe0-4cf0-bd3b-45ad1031a0c0.png)

ri0nardo:
@vernfongchao good stuff. right now we don't need the activity and plant as there isn't a feature for them yet. (activity is notifications which isnt designed, and plant is a separate feature) my trees is your profile so just clicking on your name should redirect you to the profile page. Overall it looks good.

vernfongchao:
@ri0nardo in your original design you had the email under their name, I have it removed because my email is too long so it extends past the menu, how would you want me to tackle this. would you prefer it to have something like V***@gmail.com or v...@gmail.com or just leave it out?. 

Did you want me to leave My trees up as well or remove it? currently it just redirects to the profile page same as click on name/Image

ri0nardo:
@vernfongchao lets just remove the profile image and name, and go with Profile. My trees doesn't follow the single word titles.

vernfongchao:
[...]

PR #402, opened 2023-01-14
vernfongchao/feature/Header/Redesign

vernfongchao:
Add Icons for Header

Redesign Burgur Menu and Button from https://www.figma.com/file/5C3v1LUNwMPQy9JOgnKEtr/Water-the-Trees

seprate burger menu from main header file


ri0nardo:
@vernfongchao what am i reviewing here?

vernfongchao:
@ri0nardo sorry Ill reupload images since its a different PR,  but its #366, theres some images there Ill put it on draft

vernfongchao:
> We should probably figure out a more sustainable method of importing icons than creating a new component for each icon. I would lean towards just importing them directly from mui. Are we re-using these all over the place with custom css?

the only custom css is just the font change, Theres a few icons that I consolidated into a single file based on the type of icon. I've always had a question about components like this where is it really useful for creating new components if there's a need of it being used again but with custom css other places vs just importing it again.

vernfongchao:
@zoobot @mwpark2014  made import change and added the font license to the bottom of the license page. please let me know if there's anything else.

mwpark2014:
The screenshots look amazing! Very excited about these changes - users will be delighted

client/src/components/Header/Header.scss
```diff
@@ -4,15 +4,153 @@
...
+          @media only screen and (max-width: 768px) {
+            display: flex;
+            flex-direction: column;
+            border-bottom: rgba(0, 0, 0, 0.4) 1px solid;
+          }
+
+          &__links {
+            color: black;
+
+            &:hover {
```
mwpark2014, line 127: Be careful with how heavily you nest these css selectors. This line is like 6-7 levels deep, and it seems generally accepted to keep it 3-4 levels deep (from what I can tell on https://css-tricks.com/forums/topic/sass-best-practices-nesting-more-than-3-levels-deep/).

My reasoning for not nesting too heavily: 
* It adds a bit more CSS that can bloat the CSS over time Ex: this CSS gets precompiled to `.header __navigation __authcontainer __menu __menuItem __links __links:hover`
* Adds a ton of specificity to these selectors that will make it hard to maintain over time

Another resource: https://stackoverflow.com/questions/13805324/how-bad-is-it-in-practice-to-over-nest-selectors-in-sass-scss

client/src/components/Header/Menu.js
```diff
@@ -0,0 +1,230 @@
...
+                About
+              </span>
+            </MenuItem>
+          </Link>
+
+          <Link
+            to="/contact"
+            className="header__navigation__authcontainer__menu__menuitem__links"
+          >
+            <MenuItem
```
mwpark2014, line 169: This seems like something that can be reused. Maybe have a react component like
```
const YourNewComponentName = ({ children, title }) => {
return
 <MenuItem
              sx={{
                padding: '8px',
                width: '180px',
              }}
              onClick={handleClose}
            >
              {children}
              <span className="header__navigation__authcontainer__menu__menuitem__links__span">
                {title}
              </span>
            </MenuItem>
 }
```

client/src/components/Header/Header.scss
```diff
@@ -4,15 +4,153 @@
     url('../../assets/fonts/WTT-THIN.ttf') format('truetype');
 }
 
+// .header {
```
mwpark2014, line 7: I think SASS works for this. But IIRC, you tried styled components first. IF you do want to implement this with styled components, I think this reference would help write styled components in a more concise and reusable way: https://css-tricks.com/dry-ing-up-styled-components/

Basically, styled components are designed to be reusable and modular by doing some good old-fashioned inheritance

mwpark2014 (changes requested):

client/src/pages/License/License.js
```diff
@@ -139,6 +139,85 @@ function License() {
           with respect to this CC0 or use of the Work.
         </div>
       </div>
+      <div>
+        <div className="license-p">
+          <h3>SIL OPEN FONT LICENSE Version 1.1</h3>
```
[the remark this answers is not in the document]
vernfongchao: I actually have no idea LOL

client/src/components/Section/Section.js
```diff
@@ -17,3 +18,13 @@ export default function Section({ children, title }) {
     </Box>
   );
 }
+
+export const GreenButton = styled.button((props) => ({
```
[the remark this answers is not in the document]
vernfongchao: Yeah this button is gonna be used throughout other PRs

client/src/components/Header/Header.js
```diff
@@ -1,92 +1,64 @@
...
-                <HeaderButton menuItem="User Profile" />
-              </Link>
-            </MenuItem>
-          )}
-
-          <MenuItem onClick={handleClose}>
-            <Link to="/" className="header-link">
-              <HeaderButton menuItem="Map" />
+    <div className="header">
+      <div className="header__navigation">
```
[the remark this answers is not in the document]
vernfongchao: image
I actually cant because the .header is centering the &__navigation div with different widths, I could use margins instead if it would make the nesting and redundancy a bit better.

client/src/components/Header/Menu.js
vernfongchao: good idea actually, I had trouble with the Menu not anchoring correctly to the button, but the menu items shouldnt be an issue will look into refractoring that

client/src/components/Header/Header.scss
vernfongchao: ahh interesting, I still fairly new to scss. Ill refractor the scss to keep the nesting down.

client/src/components/Header/Header.scss
vernfongchao: I'm open to having this discussion. Its just hard for me as I don't really know what conventions there are for styled components especially when it comes to using it for page structuring and being imported and being reused constantly

client/src/components/Header/Header.js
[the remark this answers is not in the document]
mwpark2014: Oh, I see. 👍

client/src/components/Header/Header.scss
mwpark2014: Okay, we can have that discussion on styled components vs SCSS on Wednesday. If we do go with styled components, we can go through some examples and establish a convention for the project

Issue #316, opened 2022-10-16
Issue with no scrolling on License Page


ri0nardo:
Link to license page - https://waterthetrees.com/license

The license page has an issue where the info below isn't accessible to certain screen sizes. I am attaching a photo of the issue from my laptop. If you zoom out you can see all the context, but not everyone can figure that out.

Issue:

Figure out why the page is not scrolling on certain screen sizes.

See issue #270 

![Image](https://user-images.githubusercontent.com/98299952/196061705-9d88f703-35f8-4401-bbc0-2d85c073cc77.png)
![Image](https://user-images.githubusercontent.com/98299952/196061704-6e382236-853a-44fe-8ec1-4bb2d6e7a91f.png)

byfar94:
Please assign it to me

CalArcher:
Can you assign this to me?

pratyushkr11:
Hello @ri0nardo I would like to work on this issue. Can you assign this to me

ri0nardo:
@pratyushkr11 have you made progress on this ticket?

Issue #315, opened 2022-10-16
Issue with no scrolling on Privacy Police Page


ri0nardo:
Link to privacy page - https://waterthetrees.com/privacy

The privacy policy page has an issue where the info below isn't accessible to certain screen sizes. I am attaching a photo of the issue from my laptop. If you zoom out you can see all the context, but not everyone can figure that out.

Issue:
- Figure out why the page is not scrolling on certain screen sizes.

![Image](https://user-images.githubusercontent.com/98299952/196061636-f9332a58-f1e8-4d0b-936b-e572e1352c12.png)

![Image](https://user-images.githubusercontent.com/98299952/196061641-5202825c-fab6-45f5-a41d-7b092030d617.png)

jazhen:
Similar issue for about, contact, license, and user profile pages. Start with the privacy page and look into the others to see if it's the same issue. If not, we can create separate issues for the other pages.

chinmaykumbhare:
hi @jazhen @ri0nardo I can fix it. Can you please assign this issue to me?

jazhen:
@chinmaykumbhare - Assigned. We found that the changes probably will affect other pages, so you will need to do some regression testing on other pages.

ri0nardo:
@chinmaykumbhare how is progress on this ticket?

Issue #212, opened 2022-03-03
Scrolling behaves strangely or not at all on the About and Contact pages


zoobot:
Make sure scroll is right and not getting cut off by footer.

fwextensions:
On desktop, there's no scrolling at all on the About and Contact pages.

adkood:
please assign it to me

zoobot:
@adkood Lisa is currently working on this.

PR #404, opened 2023-01-17
Fix:-scrolling issues #315 #316 #212

vernfongchao:
### What has Changed:
1. Overflow-hidden removed from body
2. Changed width and height from vw and vh to %

### What reviewers should know:

From what has been observed Overflow hidden was added to stop scrollbars from appearing when nothing has been overflowed; however the scrollbars appeared to do setting width and hieght to 100vw and 100vh
image

### DIFF
--- a/client/src/components/App/App.css
+++ b/client/src/components/App/App.css
@@ -49,10 +49,9 @@ body {
   text-align: left;
   padding: 0;
   margin: 0;
-  width: 100vw;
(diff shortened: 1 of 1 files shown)


zoobot (approved):
Thanks for fixing this Vern!!

Tijani-zainab:
Nice work! @vernfongchao

Issue #367, opened 2022-11-28
Implement new footer design for static screens


ri0nardo:
![Image](https://user-images.githubusercontent.com/98299952/204203846-90784fe7-58c5-48d9-82b4-c5bb9e15f481.jpg)

Redesign for footer. Current footer has scrolling and clicking issues that are currently being resolved in a different ticket. See image of current footer design below.

New footer to have even spacing and be centered to the screen. See image of both desktop and mobile footer design.

To Find the design in Figma
Open the figma link
When opened, go to the left sidebar
Find pages, Desktop
When in desktop scroll down to the about screen to find the components.

![Image](https://user-images.githubusercontent.com/98299952/204203802-7f461621-a0dd-408c-a55b-e6e53bd98277.PNG)
![Image](https://user-images.githubusercontent.com/98299952/204203801-52fbd1d1-b205-4874-8ea6-64fcaed22c83.PNG)
![Image](https://user-images.githubusercontent.com/98299952/204203803-12ca36de-cbec-45ac-96ad-1d3e49360abf.PNG)

![Image](https://user-images.githubusercontent.com/98299952/204203807-a2245552-f155-4d13-91cd-0747413a1f88.jpg)

rabbanirizwan:
@ri0nardo  kindly assign me this and also share the link of figma

ri0nardo:
@rabbanirizwan please read the notes before starting. Figma link is attached to ticket.

ri0nardo:
@rabbanirizwan how is progress on this ticket?

zoobot:
@rabbanirizwan how's this going? Still working on it?

ri0nardo:
@vernfongchao no response from assignee

ri0nardo:
@vernfongchao can we close this ticket?

vernfongchao:
@ri0nardo I am going to reopen this issue in the future and create a new pr fix, I forgot that I didn't add any margins to the bottom of the footer. I didn't add any margins because it was actually causing issues with the old design of other pages/added extra margins when the page wasn't filled up. Once the new designs of pages gets rolled in. ill add in a new pr to fix this and references this issue.

ri0nardo:
@vernfongchao do you want to backlog this ticket? & just assign yourself since you seem to have a plan.

vernfongchao:
yeah good idea! will do

ri0nardo:
[...]

PR #405, opened 2023-01-17
Feature:-Footer Redesign #367

vernfongchao:
### What has changed:
1. Add Tree Logo
2. Change font to Montserrat (wont be added in until Header redesign with the .otf file)
3. Remove positioning to only be at the bottom instead of being fixed
4. changed color to white

### Reviewer should know:
Did not add Montserrat otf file with this pr.
Design: 
image
Current PR Full Screen with hover:
image
Current PR Mobile Screen:
image

### DIFF
--- a/client/src/components/Footer/Footer.js
+++ b/client/src/components/Footer/Footer.js
@@ -1,16 +1,16 @@
 import React from 'react';
 import './Footer.scss';
 import { Link } from 'react-router-dom';
+import { TreeLogo } from '@/components/Icons';
 
 const Footer = () => (
   <div className="footer">
-    <div className="footer-content">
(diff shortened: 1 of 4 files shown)


zoobot (commented):
What's the scoop with the -- in the css? I've seen footer__content-subcontent before with BEM styles but not the --

zoobot (commented):
Overall, LGTM. Nice work!

vernfongchao:
> What's the scoop with the -- in the css? I've seen footer__content-subcontent before with BEM styles but not the -- @vernfongchao

Ohhh yeah you are absolutely correct let me change that really fast

zoobot (approved):
Nice job! @vernfongchao

PR #410, opened 2023-01-18
Replace useEffect handling tree selection in MapLayout

### LINKED ISSUE #397: For loading initial map state, replace useEffect with a callback called from onLoad

mwpark2014:
Resolves #397 

Functional change:
* No more map movements when clicking on tree markers

Non-functional changes:
* Reduced lines of code and removed a useEffect that was hard to reason about (it's called with every rerender and it's impossible differentiate initial load from url from subsequent tree selections)
* Performance gains since we're killing the hash params reset and `flyTo` side effects that used to happen on every tree selection 
* Maybe performance gains but definitely jankiness clean up from getting rid of `handleTransitionEnd`, not including the desktop tree details container in the style flow, and not creating a new dom element for every tree

Manually tested loading urls and tree selection on desktop and mobile

### DIFF
--- a/client/src/pages/Map/Map.js
+++ b/client/src/pages/Map/Map.js
@@ -202,7 +202,10 @@ export default function Map({
             hashParams.delete('id');
           }
 
-          window.location.hash = decodeURIComponent(hashParams.toString());
+          const newUrl = `${window.location.origin}#${decodeURIComponent(
+            hashParams.toString(),
+          )}`;
+          window.history.replaceState({}, '', newUrl);
         });
 
         // Unlike the click handler above, we want to get mousemove/leave events only for features
--- a/client/src/pages/Map/Map.js
+++ b/client/src/pages/Map/Map.js
@@ -2,6 +2,7 @@ import React, { useState, useEffect, useRef } from 'react';
 import mapboxgl from 'mapbox-gl';
 import { Alert, Box, Typography } from '@mui/material';
 import { useQueryClient } from 'react-query';
+import { getData } from '@/api/apiUtils';
 import { mapboxAccessToken } from '@/util/config';
 import Adopt from '@/pages/Adopt/Adopt';
 import GeolocateControl from '@/pages/UserLocation/GeolocateControl';
@@ -10,7 +11,6 @@ import { treeHealthUtil } from '@/util/treeHealthUtil';
 import MapboxControlPortal from './MapboxControlPortal';
 import MapLayers from './MapLayers';
 import TreeLayerLegend from './TreeLayerLegend';
-import { getData } from '../../api/apiUtils';
 import './Map.css';
 
 mapboxgl.accessToken = mapboxAccessToken;
@@ -143,13 +143,7 @@ export default function Map({
           url: 'mapbox://waterthetrees.open-trees',
         });
 
-        getData('trees').then((response) => {
-          const { lng, lat } = response;
-          map.flyTo({
-            center: [lng, lat],
-            zoom: 15,
-          });
-        });
+        flyToTreeAndUpdateCache(mapboxMap, queryClient, setCurrentTreeId);
 
         onLoad(mapboxMap);
         setIsMapLoaded(true);
@@ -288,4 +282,20 @@ export default function Map({
       </>
     )
   );
+}
+
+function flyToTreeAndUpdateCache(mapboxMap, queryClient, setCurrentTreeId) {
+  const currentTreeId = new URLSearchParams(window.location.hash.slice(1)).get(
+    'id',
+  );
+  const queryKey = ['trees', { id: currentTreeId }];
+  getData({ queryKey }).then((response) => {
+    const { lng, lat } = response;
+    mapboxMap.flyTo({
+      center: [lng, lat],
+      zoom: 15,
+    });
+    queryClient.setQueryData(queryKey, response);
+    setCurrentTreeId(currentTreeId);
+  });
 }
--- a/client/src/pages/Map/Map.js
+++ b/client/src/pages/Map/Map.js
@@ -10,6 +10,7 @@ import { treeHealthUtil } from '@/util/treeHealthUtil';
 import MapboxControlPortal from './MapboxControlPortal';
 import MapLayers from './MapLayers';
 import TreeLayerLegend from './TreeLayerLegend';
+import { getData } from '../../api/apiUtils';
 import './Map.css';
 
 mapboxgl.accessToken = mapboxAccessToken;
@@ -140,6 +141,14 @@ export default function Map({
         mapboxMap.addSource('WTTV', {
           type: 'vector',
           url: 'mapbox://waterthetrees.open-trees',
+        });
+
+        getData('trees').then((response) => {
+          const { lng, lat } = response;
+          map.flyTo({
+            center: [lng, lat],
+            zoom: 15,
+          });
         });
 
         onLoad(mapboxMap);
--- a/client/src/pages/Map/Map.js
+++ b/client/src/pages/Map/Map.js
@@ -2,7 +2,7 @@ import React, { useState, useEffect, useRef } from 'react';
 import mapboxgl from 'mapbox-gl';
 import { Alert, Box, Typography } from '@mui/material';
 import { useQueryClient } from 'react-query';
-import { getData } from '@/api/apiUtils';
+import { useTreeQuery } from '@/api/queries';
 import { mapboxAccessToken } from '@/util/config';
 import Adopt from '@/pages/Adopt/Adopt';
 import GeolocateControl from '@/pages/UserLocation/GeolocateControl';
@@ -86,6 +86,17 @@ export default function Map({
   const queryClient = useQueryClient();
   const selectionEnabledRef = useRef(selectionEnabled);
   const currentFeature = useRef(null);
+  const hasInitialFlyToFired = useRef(false);
+
+  const initialLoadTreeId = useRef(
+    new URLSearchParams(window.location.hash.slice(1)).get('id'),
+  );
+  // This tree query is intentionally separate from the one in `MapLayout`,
+  // since it kicks off only once on load!
+  const { data: initialLoadTreeData } = useTreeQuery(
+    { id: initialLoadTreeId.current },
+    { retry: 0, enabled: isMapLoaded && !!initialLoadTreeId },
+  );
 
   // TODO: maybe use a class for Map so that event handlers bound to the instance can look at
   //  this.props.selectionEnabled instead of having to mirror it in a ref like this
@@ -142,8 +153,6 @@ export default function Map({
           type: 'vector',
           url: 'mapbox://waterthetrees.open-trees',
         });
-
-        flyToTreeAndUpdateCache(mapboxMap, queryClient, setCurrentTreeId);
 
         onLoad(mapboxMap);
         setIsMapLoaded(true);
@@ -250,6 +259,17 @@ export default function Map({
     }
   }, [map, containerRef]);
 
+  useEffect(() => {
+    if (!isMapLoaded || hasInitialFlyToFired.current) {
+      return;
+    }
+    if (initialLoadTreeData) {
+      const { lng, lat, id } = initialLoadTreeData;
+      flyTo({ map, lng, lat, hasInitialFlyToFired });
+      setCurrentTreeId(id);
+    }
+  }, [isMapLoaded, map, initialLoadTreeData, setCurrentTreeId]);
+
   if (!isMapboxSupported) {
     return unsupportedError;
   }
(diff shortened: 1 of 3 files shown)


client/src/pages/Map/Map.js
```diff
@@ -2,6 +2,7 @@ import React, { useState, useEffect, useRef } from 'react';
 import mapboxgl from 'mapbox-gl';
 import { Alert, Box, Typography } from '@mui/material';
 import { useQueryClient } from 'react-query';
+import { getData } from '@/api/apiUtils';
```
zoobot, line 5: I think we should not bypass react-query and call the apiUtils on their own as react-query does caching, etc. The react-query obfuscation in api/queries is not easy to read unfortunately. If you need to bypass that and use react-query on it's own, I think that's a better option. I've been meaning to refactor api/queries for some time. https://react-query-v3.tanstack.com/ 
We've discussed moving graphQL or redux at times as it seems that react-query itself is not super easy to understand.

client/src/pages/Map/Map.js
```diff
@@ -277,3 +283,21 @@ export default function Map({
     )
   );
 }
+
+function flyToTreeAndUpdateCache(mapboxMap, queryClient, setCurrentTreeId) {
```
zoobot, line 287: Prefer async await syntax over then for code styling.

Here's documentation for using https://react-query-v3.tanstack.com/quick-start if you want to start bringing this back in to the main code base instead of using the api/queries.

zoobot (changes requested):
The map's not jumping and zooming around anymore is great. Definitely make this more pleasant to use!! Thanks!!

Request a few changes/also added more comments in the code that may be worth more discussion about react-query and the api/queries in general.

Can you move the geolocation and navigation controls to the left so they are not covered by the drawer with this PR changes? thanks!

These small changes:

`mapboxMap.addControl(new mapboxgl.NavigationControl(), 'top-left');`

And this too where it says top-right
` <MapboxControlPortal map={map} position="top-left">`

client/src/pages/Map/Map.js
mwpark2014: Yup, that's exactly the main point of contention for this PR! I know why we use `react-query` - it provides a lot of useful reusable logic related to making queries such as error handling and caching, which is great. I think we should keep using it rather than moving on to something else.

However, using `react-query` takes away control from us in this specific situation. I noticed that there's a bunch of `useEffect`s in `MapLayout` even though using `react-query` is supposed to _reduce_ the need to use the `useEffect` hook.  That's because we make a tree query as soon as `MapLayout` mounts. Now, we must wait for both 1) `react-query` to wait for a response and update the state of `MapLayout` and 2) `Map` to kick off the `mapboxMap.load` callback which also updates the state of `MapLayout`. Now, we need a `useEffect` to kick off when we have both the tree query response and map load, which was buggy in the previous implementation, and is hard to implement in this iteration in a way where it'll be free from bugs in the future.

Before, we combined two updates 1) Kick off a map update side effect when the page loads with a certain tree id and 2) Kick off a map update side effect when a tree id is changed. It kind of worked before when we conflated both. Now, we want to separate map load and tree selection behaviors.

Maybe I don't quite know what you had in mind for "bypassing api/queries and using react-query". What I need is a piece of asynchronous code related to making tree queries that I can `await` in the `mapboxMap.load` callback. `react-query` does not provide that since it abstracts asynchronous code into _React lifecycle state_. I'm open to any alternate suggestions!

Here are some example alternative approaches I tried with `react-query`  to get it to just fire **once** upon page load that doesn't quite work:
* Only call the `useEffect` hook once upon mounting `MapLayout`. This doesn't work since we need to wait for the response of the tree query
* Have the `useEffect` hook be called every time `map` and `currentTreeDb` is updated, but keep some kind of flag that signifies the "first" time and only call it if it's the "first" time and the map is loaded. Doesn't work because map doesn't necessarily load with a tree id in the URL.
* Use the `react-query` `onSuccess` callback. This is called every time the tree query is fetched, including refetches.

Instead, with these changes, we avoid calling `react-query` in this instance here. We then update the `react-query` cache manually to avoid any issues with double calls. We depart from the `react-query` standard in this one instance. However, I would argue the code is much easier to read and reason about. There is a strong reason for why we should do it this way.

client/src/pages/Map/Map.js
mwpark2014: Oops, I don't know why I reverted back to old fashioned promise chaining lol. 👍

client/src/pages/Map/Map.js
zoobot: I feel your pain and thank you for trying a bunch of different ways. I'm am glad you want to keep with react-query as I think it's pretty nice. It is much easier to read.

Have you tried a lazy query? 
I have not tested this code much...
`
 const queryKey = ['trees', { id: currentTreeId }];
  const { data } = useQuery({
    queryKey,
    queryFn: () => getData({ queryKey }),
    // ⬇️ disabled as long as the currentTreeId is empty
    enabled: !!currentTreeId,
    // call flyto function on settled?
    onSettled: (data) => {
      flyToTreeAndUpdateCache(
        map,
        queryClient,
        setCurrentTreeId,
        data,
        currentTreeDataVector, 
      );
    },
    // If set to true, the query will refetch on mount if the data is stale.
    refetchOnMount: true,
  });`
  
currentTreeId would need to get passed from map

react-query has a lot of cool options to deal with handling refetch, etc. I haven't tried them all but added some that could possibly be used in this case.
https://tanstack.com/query/latest/docs/react/guides/important-defaults?from=reactQueryV3&original=https%3A%2F%2Freact-query-v3.tanstack.com%2Fguides%2Fimportant-defaults


https://tanstack.com/query/latest/docs/react/reference/useQuery?from=reactQueryV3&original=https%3A%2F%2Freact-query-v3.tanstack.com%2Freference%2FuseQuery#:%7E:text=staleTime%3A%20number%20%7C%20Infinity

https://tanstack.com/query/latest/docs/react/guides/disabling-queries?from=reactQueryV3&original=https%3A%2F%2Freact-query-v3.tanstack.com%2Fguides%2Fdisabling-queries#lazy-queries



One thing I noticed is this doesn't handle the vector data /vector tree sharing of hash params(our current prod version doesn't either). 
Jason started that but until it's working we are removing the id in prod now. (I believe he's no longer working on it)
Some of your cache function could get that to work I think. Are you interested in working on getting that working while you are dealing with this part of the code base? If so, we let's put that in a new branch.
I added a bunch of links that could potentially help. 

The way we are currently combining db and vector data could definitely use a refactor as it's become spaghetti in MapLayout.js and Map.js

https://docs.mapbox.com/mapbox-gl-js/api/map/#map#loaded

https://docs.mapbox.com/mapbox-gl-js/example/filter-features-within-map-view/

https://github.com/mapbox/mapbox-gl-js/issues/4222#issuecomment-279446075

mwpark2014: Yup, I tried a lazy query. The problem is that it calls `onSettled` every time there's a refetch, which is a core feature of `react-query`, and also it runs every single time `currentTreeId` changes.

Addressing each resource:
* Important defaults - yup, played around with this, which spawned https://github.com/waterthetrees/wtt_front/pull/413. 
* `useQuery` API - yup, I experimented with various status state and callbacks
* Lazy queries - In addition to what has already been said, this would require some sort of lock, and I could only think of v complicated approaches

> One thing I noticed is this doesn't handle the vector data /vector tree sharing of hash params(our current prod version doesn't either).
From what was demonstrated at the demo, it does handle vector data. There was a bug at the time you probably reviewed where a sidebar did not appear if no tree data existed for the specific tree, but I pushed a commit to fix that

zoobot: What about setting enabled state where you currently have the flyToTreeAndUpdateCache call in onload, or possibly use memo to take advantage of https://reactjs.org/docs/react-component.html#shouldcomponentupdate?

I know this is a bit of a nitpick, I just don't want to introduce a workaround if we can find another solution as our current workarounds have added tech debt like a year long. Thanks for your patients.

```
const [isEnabled, setIsEnabled] = useState()
const queryKey = ['trees', { id: currentTreeId }];
  const { data } = useQuery({
    queryKey,
    queryFn: () => getData({ queryKey }),
    // ⬇️ disabled as long as the isEnabled is false
    enabled: isEnabled,
    // call flyto function on settled?
    onSettled: (data) => {
      flyToTreeAndUpdateCache(
        map,
        queryClient,
        setCurrentTreeId,
        data,
        currentTreeDataVector, 
      );
    },
    // If set to true, the query will refetch on mount if the data is stale.
    refetchOnMount: true,
  });
  ```

Comment (mwpark2014) [reply client/src/pages/Map/Map.js:5]: Okay, here's something we can do in pseudocode:
```
If (tree ID in the URL) {
  isEnabled = true;
}
onSettled: (data, error) => {
  if (!error && map.loaded && isEnabled) {
    flyToTree([data.lng, data.lat]);
  }  else if (!error && isEnabled) {
    setQueuedUpFlyTo([data.lng, data.lat]) // If the map takes longer to load than the tree query
  }
  isEnabled = false;
}

in Map.js
map.load(() => {
  if (queuedUpToFlyTo) {
   flyToTree(queuedUpToFlyTo);
   isEnabled = false;
  }
});
```
^ I think even this has a race condition bug in it where `onSettled` can be called right after `map.load` and before the state change rerenders the components. I also don't like how the logic is spread out between multiple components, and I think it'll be even harder to understand when we add the geolocation check.  

I understand the reluctance to diverge from using one pattern everywhere, but I do think this is a case that `react-query` is ill-suited for until a cleaner solution is found. We shouldn't be using `react-query` for use cases where it makes things less understandable.

My argument is:
My solution in the PR is cleaner than the useEffect() that listened for changes to map, currentTreeDb, and currentTreeDataVector state changes implementation. It's encapsulated in a helper function that is called within `map.load` that is actually much easier to update. I will have to add some extra logic here when we add flying to geolocation upon load, and I think future maintainability won't be bad.

`react-query` abstracts asynchronous calls into react lifecycles and that's counterproductive when we want to only make a call as a side effect of an asynchronous call. The lazy queries are good when we want to make a call as a side effect of another `react-query` call or react state change.

mwpark2014: Ooh, I thought of a much better way of doing it based on your suggestions. Give me time to make a commit with that approach since I'll need to take some time to think about how it'll fit in with other future onLoad side effects we want to introduce - not sure I like this approach yet but it'll keep the use of `react-query`

zoobot: Nice, thanks for giving it thought! It's awesome you are trying to make it easier to read, slim down the code, and have a goal of performance gains. 

Totally agree that spreading the functionality across multiple components makes for a bit of spaghetti reading. If we can clean that up some, it'll be a lot easier to jump in and work on that part of the code.

I also was checking out dependent queries that could potentially be used with the enable.
https://tanstack.com/query/latest/docs/react/guides/dependent-queries?from=reactQueryV3&original=https%3A%2F%2Freact-query-v3.tanstack.com%2Fguides%2Fdependent-queries

Let me know how it goes today.

zoobot: @mwpark2014 How'd it go? 
Do you still prefer bypassing react-query for this or do you like your new idea better? 

It does make the code cleaner and makes the site easier to use so would love to get this into prod.

mwpark2014: Yeah, sorry. Planning on getting an iteration out today. This is taking a ton of thinking lol in terms of what is the best way for future maintainability

mwpark2014: @zoobot 
Okay, here's the alternate solution with `react-query`. My mind block was separating updates from subsequent tree clicks from the initial tree query. In this solution, we have two separate tree queries - one in `Map` and one in `MapLayout`. I still don't love this solution because I think could it be slightly easier to introduce a bug - i.e., this useEffect is called every time the component updates unlike the solution in this PR which is guaranteed to only fire once on map load. 

But I admit that it's concise and keeps to the "use react-query everywhere" standard. I'm okay with going with either. Your call @zoobot!

https://github.com/waterthetrees/wtt_front/pull/422/files#diff-9d7e66879ba8b56e3e2be3dfd37cb4655ef5cd17188ca9b82d0301381514d6fe

mwpark2014: If we go with the other solution, let me merge the commit into this PR instead of merging the other draft PR so we can retain all the comments

mwpark2014:
@zoobot I merged in the commit associated with the alt-demo and rebased off of main. Could you approve this PR and merge at your convenience?

zoobot (approved):
Nice work!! :) Glad to get this in!!

mwpark2014:
> Hey so just doublechecking. I generally squash and merging which will squash all your commits. That ok or do you want to do the squashing manually for cleanup and I can rebase merge? @mwpark2014

I squashed and cleaned up the commit messages!

PR #413, opened 2023-01-20
Update default react query stale time to 5 min

mwpark2014:
The docs _say_ that the cache query keys should be invalidated within 5 min... but from my testing, it doesn't appear that we are caching queries at all.

Which is a shame, since @fwextensions took the time to write this:
https://github.com/waterthetrees/wtt_front/blob/6bca4d3951e2e6fb7d2a8944f8399a52c40c4332/client/src/pages/Map/Map.js#L177-L185

Before:
image

After:
image

Only diff is the disappearance of the `State` field: 
image
That's probably something missing from the geojsons for some reason

### DIFF
--- a/client/src/pages/Map/Map.js
+++ b/client/src/pages/Map/Map.js
@@ -175,18 +175,6 @@ export default function Map({
             };
 
             setCurrentTreeDataVector(currentTree);
-            const queryKeys = ['trees', { id }];
-
-            // Cache the properties from the vector tile as the data for the /trees query that will
-            // be triggered by the setCurrentTreeId() call below.  The Tree will first
-            // get this cached data and then update it when the server response comes back.  But
-            // only do this if there's no cached data already.  If there is, that data is presumably
-            // the latest response from the server, so we don't want to override it with older data
-            // from the vector tile.
-            if (!queryClient.getQueryState(queryKeys)) {
-              queryClient.setQueryData(queryKeys, properties);
-            }
-
             setCurrentTreeId(id);
             hashParams.set('id', id);
             mapboxMap.getCanvas().style.cursor = 'pointer';
--- a/client/src/pages/Map/MapLayout.js
+++ b/client/src/pages/Map/MapLayout.js
@@ -26,7 +26,6 @@ const MapContainer = styled('main')({
 function MapLayout() {
   const [currentTreeId, setCurrentTreeId] = useState();
   const [currentTreeDataVector, setCurrentTreeDataVector] = useState();
-  const [currentTreeData, setCurrentTreeData] = useState();
   const [map, setMap] = useState(null);
   const [mapSelectionEnabled, setMapSelectionEnabled] = useState(true);
   const { newTreeState } = useNewTree();
@@ -53,14 +52,10 @@ function MapLayout() {
     setMapSelectionEnabled(!newTreeState.isDragging);
   }, [newTreeState.isDragging]);
 
-  useEffect(() => {
-    if (currentTreeDb) {
-      setCurrentTreeData({
-        ...currentTreeDb,
-        ...currentTreeDataVector,
-      });
-    }
-  }, [currentTreeDb]);
+  const currentTreeData = {
+    ...currentTreeDataVector,
+    ...currentTreeDb,
+  };
 
   return (
     <Box sx={{ display: 'flex' }}>
--- a/client/src/components/App/App.js
+++ b/client/src/components/App/App.js
@@ -33,6 +33,7 @@ const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       refetchOnWindowFocus: false,
+      staleTime: 1000 * 60 * 5, // 5 minutes
     },
   },
 });


zoobot (commented):
Please make sure caching changes works on any buttons, inputs, etc that has mutations in the Tree and AddTree components. 
https://react-query-v3.tanstack.com/guides/query-invalidation
On doing tree maintenance
1. Make sure the data changes and is persistent when clicking another tree and then the same tree again
2. Make sure the data changes when copying to new page
3. When changing tree health, make sure dot changes color on the map.
4. When adding a tree, make sure it shows on the map and is persistent on refresh

Test on desktop and mobile.

zoobot (dismissed):
LGTM Nice catch! Is this ready to go?

mwpark2014:
> Please make sure caching changes works on any buttons, inputs, etc that has mutations in the Tree and AddTree components. https://react-query-v3.tanstack.com/guides/query-invalidation On doing tree maintenance
> 
> 1. Make sure the data changes and is persistent when clicking another tree and then the same tree again
> 2. Make sure the data changes when copying to new page
> 3. When changing tree health, make sure dot changes color on the map.
> 4. When adding a tree, make sure it shows on the map and is persistent on refresh
> 
> Test on desktop and mobile.

1 and 3 tested successfully.
Need clarification on what 2 means.
For 4) I run into 
image
but I also run into that for dev.waterthetrees.com as well

mwpark2014:
Hmm, looks like I'm losing data when I click on a tree... I wonder if it's connected to https://github.com/waterthetrees/wtt_front/blob/6bca4d3951e2e6fb7d2a8944f8399a52c40c4332/client/src/pages/Map/Map.js#L177-L185...

image

mwpark2014:
> LGTM Nice catch! Is this ready to go?

Nope, not ready. Idk how to mark the PR as needs work, but I need to follow up on a couple things

mwpark2014:
@zoobot do you have any insight to why trees that have been edited are stripped of lots of their properties in the vector data?

Edited tree feature properties (upon map click):
image

Unedited tree feature properties (upon map click):
image

This will probably block this PR from being merged since activating react-query caching causes a regression.

Due to https://github.com/waterthetrees/wtt_front/blob/6bca4d3951e2e6fb7d2a8944f8399a52c40c4332/client/src/pages/Map/Map.js#L177-L185

From what I understand, old way was: 
User clicks on tree feature on map => Vector data is loaded into cache (not sure why - to be available immediately in the tree details container?) => Tree query returns a response which then overwrites the data since the data is considered immediately stale (so tree query is source of truth)

New way:
User clicks on tree feature on map => Vector data is loaded into cache => Tree query never runs since cache is populated and value is not stale. 
If it's not easy to keep tree vector data completely hydrated with our tree db data, then I'm not even sure if it makes to have this intermediary cache load step. Alternatively, I'm not sure if Tom's work would keep these vector data in sync with the db.

mwpark2014:
Yeah, if the intention is to just combine vector data and treedb treedata then we shouldn't be doing this cache update logic in 
```
 if (!queryClient.getQueryState(queryKeys)) { 
   queryClient.setQueryData(queryKeys, properties); 
 } 
```

`currentTreeDb` should map to only treedb data. This cache update logic attempts to combine vector data into `currentTreeDb`, which will produce unintentional results imo.
I don't think we need to have a custom hook either. Couldn't we just calculate a `const combinedTreeData = currentTreeDb || currentTreeDataVector` and pass that value into child components instead of doing anything fancy? Seems unnecessary to have this additional `currentTreeData` state that triggers an extra rerender. 

I think changing this logic belongs in this PR so that it unblocks increasing `react-query`'s stale time, so I'll include that small refactor in this PR

client/src/pages/Map/MapLayout.js
```diff
@@ -53,14 +52,10 @@ function MapLayout() {
...
 
-  useEffect(() => {
-    if (currentTreeDb) {
-      setCurrentTreeData({
-        ...currentTreeDb,
-        ...currentTreeDataVector,
-      });
-    }
-  }, [currentTreeDb]);
+  const currentTreeData = {
```
mwpark2014, line 55: Just a small change - this doesn't need to be in state since it only updates when `currentTreeDataVector` or `currentTreeDb` updates

client/src/pages/Map/Map.js
```diff
@@ -175,18 +175,6 @@ export default function Map({
...
             setCurrentTreeDataVector(currentTree);
-            const queryKeys = ['trees', { id }];
-
-            // Cache the properties from the vector tile as the data for the /trees query that will
-            // be triggered by the setCurrentTreeId() call below.  The Tree will first
-            // get this cached data and then update it when the server response comes back.  But
-            // only do this if there's no cached data already.  If there is, that data is presumably
-            // the latest response from the server, so we don't want to override it with older data
-            // from the vector tile.
-            if (!queryClient.getQueryState(queryKeys)) {
```
mwpark2014, line 186: This is actually completely unnecessary. We are combining vector data and tree db data correctly in `MapLayout` as it was

mwpark2014:
@zoobot This PR is ready for review. Not many changes needed actually (thankfully)

client/src/pages/Map/Map.js
zoobot: nice catch

client/src/pages/Map/MapLayout.js
zoobot: awesome nice work

zoobot (approved):

mwpark2014:
@zoobot oh yeah, my bad. I should have specified that this is meant to be merged into main _after_ the other PR since it depends on it. (sorry, I'm used to hitting the merge button myself) The PR was into the other branch since this PR is focused on only 2 commits out of the 14 or so altogether.

Should be easy to fix - I'll create another PR to replace this one with the cherry-picked commits

mwpark2014:
> sounds good, we should also revert the merge into useEffect branch as I merged it in there thinking it was pointed to main. cherrypicking doesn't remove the commit. Let me know if you want me to do anything.

Yup, yup. Already done :)

Issue #411, opened 2023-01-19
Remove zoom from mobile


ri0nardo:

WAChattha05:
I want to solve this issue please assign me

PR #414, opened 2023-01-20
Fixed mobile zoom control

GabrielBorgess:
removed the zoom control buttons for mobile devices since we already pinch the screen to do that.
#411


Tijani-zainab:
Hi @GabrielBorgess, Well-done
was there a ticket for this solution?

PR #417, opened 2023-01-24
move geolocate and navigation controls left to accomodate drawer

zoobot:


mwpark2014:
Oops, good catch on the geolocate and navigation controls being overridden. However, I'm not sure if it's intuitive to have the geolocate and navigation controls on the left. Calls to action are usually on the right hand side in my experience just by convention, what do you think @ri0nardo?

I don't mind making the changes to prevent the sidebar from hiding the controls in the original PR.

ri0nardo:
> Oops, good catch on the geolocate and navigation controls being overridden. However, I'm not sure if it's intuitive to have the geolocate and navigation controls on the left. Calls to action are usually on the right hand side in my experience just by convention, what do you think @ri0nardo? (at the very least, that's where the map controls usually are for maps I think)
> 
> I don't mind making the changes to prevent the sidebar from hiding the controls in the original PR.

Currently how the site is laid out it makes sense to have it on the left, but if we are planning on moving the drawer then we would move it to the right side. It would make sense to have it on the right as the mobile design would have buttons and actions closer to right side of a device.

mwpark2014:
Okay, I'm alright with that plan! Just to make sure I understand. The plan is to:

1) Move controls to the left
2) Merge in the changes related https://github.com/waterthetrees/wtt_front/pull/410 to once approved
3) Separate PR for moving the controls to the right and the drawer to the left (moving drawer to the left should be easy after 2) I believe!)

Do I have that right @zoobot ?

mwpark2014 (approved):

Tijani-zainab:
Nice work! @zoobot 
looks likes it settled to move the controls to the left!

PR #420, opened 2023-01-30
Carbon stats added to sidebar

zoobot:
I updated the [old carbon calc branch](https://github.com/waterthetrees/wtt_front/tree/zoobot/feature/carbon-calculator) to get us going on this. This is just a start and not ideal as so much info is missing but it can be merged as-is. 
 
 Note: Many cities don't have age/height
 I just defaulted age/height to 1 if we don't have that info. Not having it screws up the calcs. 
 
 If the Davey gives us a calculations list for our DB or we do some api calls to Davey, we can get better maths and move all the calculation to the backend.

### DIFF
--- a/client/src/pages/Tree/CarbonCalculator.js
+++ b/client/src/pages/Tree/CarbonCalculator.js
@@ -1,0 +1,222 @@
+/* eslint-disable max-len */
+/* eslint-disable react/jsx-one-expression-per-line */
+import React from 'react';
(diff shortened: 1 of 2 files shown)


Tijani-zainab:
Nice work @zoobot 
Nice work for displaying the results  in a table along with an explanation of each calculation step!

Tijani-zainab (approved):
Nice work!

PR #422, opened 2023-01-31
Replace use effect alt demo

mwpark2014:
Alternate solution to https://github.com/waterthetrees/wtt_front/pull/410

### DIFF
--- a/client/src/pages/Map/Map.js
+++ b/client/src/pages/Map/Map.js
@@ -202,7 +202,10 @@ export default function Map({
             hashParams.delete('id');
           }
 
-          window.location.hash = decodeURIComponent(hashParams.toString());
+          const newUrl = `${window.location.origin}#${decodeURIComponent(
+            hashParams.toString(),
+          )}`;
+          window.history.replaceState({}, '', newUrl);
         });
 
         // Unlike the click handler above, we want to get mousemove/leave events only for features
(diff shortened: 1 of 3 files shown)


client/src/pages/Map/Map.js
```diff
@@ -85,6 +86,17 @@ export default function Map({
   const queryClient = useQueryClient();
   const selectionEnabledRef = useRef(selectionEnabled);
   const currentFeature = useRef(null);
+  const hasInitialFlyToFired = useRef(false);
```
mwpark2014, line 89: This is not critical right now, but it will be in the future when we add other onLoad flyTo effects such as geolocation. We want to make sure we don't fly to user's initial tree Id and then flies to user's geolocation, etc.

client/src/pages/Map/Map.js
```diff
@@ -85,6 +86,17 @@ export default function Map({
...
+
+  const initialLoadTreeId = useRef(
```
mwpark2014, line 91: It's pretty key that this is in a `useRef`. This value should NOT change on any rerender since we don't want this id to respond to updates to the URL params

zoobot (commented):
@mwpark2014 Thanks, this looks more react-like. Your other version was maybe easier to read but I think with the nice variable names in this, we can avoid having to comment the code, this is good and let's use it! We should keep the other version around in case we want to revert or this doesn't work out.

zoobot (approved):

mwpark2014:
> @mwpark2014 Thanks, this looks more react-like. Your other version was maybe easier to read but I think with the nice variable names in this, we can avoid having to comment the code, this is good and let's use it! We should keep the other version around in case we want to revert or this doesn't work out.

Sure, I can get behind this. Let's 🚀 !

mwpark2014:
> Feel free to merge this.

@zoobot I merged this branch into https://github.com/waterthetrees/wtt_front/pull/410. Could you approve that instead so we provide an easy link to all the relevant PR discussion and context from the merge commit?

PR #429, opened 2023-02-02
Fix for invisible drawer

mwpark2014:


zoobot (approved):
thanks for quick fix!

PR #511, opened 2023-03-02
Update snapshot and fix error

mwpark2014:
This snapshot was auto-updated with `npm test -- -u`. 

I also fixed a console warning that the snapshot test notified me about


zoobot (approved):
LGTM, thanks for adding in the carbon test fix!

PR #512, opened 2023-03-02
Improve snapshot tests

### LINKED ISSUE #496: Link up jest tests with github CI
We just added a couple of jest tests to the codebase. We need to link them to Github CI to have them automatically run on every PR

mwpark2014:
Resolves #496 

* Self-documenting comment on how to update snapshots
* Add jest tests to CI
* Move mocks out of the root directory to be a little cleaner
* Reuse query data mocking automatically for all jest tests


client/src/tests/setup.js
```diff
@@ -0,0 +1,67 @@
+/**
```
mwpark2014, line 1: This file will be run before all tests automatically and is intended to be a convenient way to reuse query mocking for all tests. I'm a little worried that it'll happen a little too magically and devs won't know where the test data is coming from, but hopefully developers will find this file and see this comment.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,18 @@
...
+    branches:
+      - main
+  pull_request:
+    branches: [main]
+
+jobs:
+  Jest:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
```
zoobot, line 14: Nice!! Glad to have a PR action for test running!!

client/src/tests/setup.js
zoobot: Maybe put a BOLD comment at the top of the readme or create a template test that has a comment on the top?

zoobot (approved):
LGTM, left a few comments.

client/src/tests/setup.js
zoobot: The name does make it pretty obvious so its probably fine as-is

mwpark2014: Updated the readme!

Issue #523, opened 2023-03-05
Hook up search bar to API call and mapbox camera action


mwpark2014:
Dependent on https://github.com/waterthetrees/wtt_server/issues/130

* Make the actual API calls for address and potentially places and figure out where this logic should live
* Parse direct lat/long search queries
* Make the map pan and zoom into the target location
* Debounce search requests to save on # of API calls
* Potentially cache most recent searches
* Error handling, empty search results

Issue #522, opened 2023-03-05
Implement search bar UI component on mobile/desktop


mwpark2014:
Finalized mocks TBD but discussion here: https://github.com/waterthetrees/wtt_server/issues/120

https://www.figma.com/file/5C3v1LUNwMPQy9JOgnKEtr/Water-the-Trees?node-id=2700-20928

Mobile/desktop likely to have differences (mobile search results should take up whole screen with a dialog)

* Input field
* Autocomplete menu, which replaces input field text when result is selected
* Search results organized into categories
* Custom styled search result rows
* Keyboard shortcuts (Split up into separate github issue if not out of the box)
* Accessibility (Split up into separate github issue if not out of the box)

See MUI components

mwpark2014:
@ri0nardo It looks like you removed the searchbar mock from the figma. Any chance you can add it back or link me to it elsewhere? It was a good visual to base my first iteration component on

ri0nardo:
@mwpark2014 the design is still on figma, I just organized the file so everything is categorized. To find the design, go to the left side panel where pages are and click on "search / side menu design"
![Capture](https://user-images.githubusercontent.com/98299952/226091640-760b0113-3351-4ee5-b8e0-6e6e40d719b5.PNG)

mwpark2014:
> @mwpark2014 the design is still on figma, I just organized the file so everything is categorized. To find the design, go to the left side panel where pages are and click on "search / side menu design" ![Capture](https://user-images.githubusercontent.com/98299952/226091640-760b0113-3351-4ee5-b8e0-6e6e40d719b5.PNG)

oh, my bad. that's awesome!

PR #573, opened 2023-03-19
Searchbar Input Field, Autocomplete, and SearchResult UI components

mwpark2014:
Resolves #522 

The UI component is just being piped in mock data for now. Next PR will be funneling actual search results in

Screenshots:
Desktop:
image
Mobile:
image

PR #574, opened 2023-03-19
Add ability to make mapbox geocoding API requests and execute mapbox camera actions

mwpark2014:
First iteration of the map search! Lots of things to improve, but I think it's pretty nifty even with what it is right now. 

Resolves https://github.com/waterthetrees/wtt_front/issues/523 and depends on https://github.com/waterthetrees/wtt_front/issues/522

Only pans the map, doesn't zoom, since that seems consistent with behavior in https://docs.google.com/document/d/1MSSbkCo77VCE9tHZdWKAr8lfrBpRF_1AEMptInFjUBI/edit#

https://user-images.githubusercontent.com/10445556/226207959-74ce48a0-4e19-4c93-9950-5e41368e3ddd.mp4

### DIFF
--- a/client/src/components/Header/Header.js
+++ b/client/src/components/Header/Header.js
@@ -11,8 +11,7 @@ const Header = () => {
   const isMobile = useMediaQuery('(max-width:768px)');
   const location = useLocation();
   const isMapPage = location.pathname == '/' || location.pathname == '/map';
-  // FIXME when search feature is enabled
-  const showHeaderLink = true || !isMobile || !isMapPage;
+  const showHeaderLink = !isMobile || !isMapPage;
 
   const handleClick = (event) => {
     setAnchorEl(event.currentTarget);
--- a/client/src/pages/Search/Search.js
+++ b/client/src/pages/Search/Search.js
@@ -1,5 +1,6 @@
 import React, { useState } from 'react';
 import { useQuery } from 'react-query';
+import { debounce } from '@mui/material/utils';
 import SearchBar from '@/components/Search/SearchBar';
 import { mapboxAccessToken } from '@/util/config';
 
@@ -8,6 +9,7 @@ import './Search.scss';
 const MIN_CHARS_FOR_SEARCH = 2;
 const BASE_MAPBOX_SEARCH_API_URL =
   'https://api.mapbox.com/geocoding/v5/mapbox.places/';
+const SEARCH_QUERY_CACHE_TIME = 1000 * 30; // 30 seconds
 
 const Search = () => {
   const [query, setQuery] = useState('');
@@ -15,9 +17,13 @@ const Search = () => {
     queryKey: query,
     queryFn: () => getSearchResults(query),
     enabled: query.length > MIN_CHARS_FOR_SEARCH,
+    cacheTime: SEARCH_QUERY_CACHE_TIME,
   });
+
+  // Debounce search reqeusts to mitigate churning through our API requests budget
+  const debouncedSetQuery = debounce((query) => setQuery(query), 250);
   const handleChange = (event) => {
-    setQuery(event.currentTarget.value);
+    debouncedSetQuery(event.currentTarget.value);
   };
 
   const options = searchResults?.features.map((result) => ({
--- a/client/src/pages/Search/Search.js
+++ b/client/src/pages/Search/Search.js
@@ -11,7 +11,7 @@ const BASE_MAPBOX_SEARCH_API_URL =
   'https://api.mapbox.com/geocoding/v5/mapbox.places/';
 const SEARCH_QUERY_CACHE_TIME = 1000 * 30; // 30 seconds
 
-const Search = () => {
+const Search = ({ map }) => {
   const [query, setQuery] = useState('');
   const { data: searchResults, isLoading } = useQuery({
     queryKey: query,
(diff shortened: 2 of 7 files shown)


client/src/pages/Search/Search.js
```diff
@@ -0,0 +1,88 @@
...
+    isError,
+    isSuccess,
+  } = useQuery({
+    queryKey: query,
+    queryFn: () => getSearchResults(query),
+    enabled: query.length > MIN_CHARS_FOR_SEARCH,
+    cacheTime: SEARCH_QUERY_CACHE_TIME,
+  });
+
+  // Debounce search reqeusts to mitigate churning through our API requests budget
```
zoobot, line 28: requests :D

client/src/pages/Search/Search.js
```diff
@@ -0,0 +1,88 @@
...
+  const debouncedSetQuery = debounce((query) => setQuery(query), 250);
```
zoobot, line 29: Great idea

client/src/pages/Search/Search.js
```diff
@@ -0,0 +1,88 @@
...
+      />
+    </div>
+  );
+};
+
+const getSearchResults = async (query) => {
+  const hashParams = new URLSearchParams(window.location.hash.slice(1));
+  const coords = hashParams.get('pos')?.split('/');
+  const searchParams = new URLSearchParams({
+    access_token: mapboxAccessToken,
```
zoobot, line 71: Our access token already has permissions for reverse geolocation?

client/src/pages/Search/Search.js
```diff
@@ -0,0 +1,88 @@
...
+    autocomplete: true,
+    fuzzymatch: true,
+    proximity: `${coords[2]},${coords[1]}`,
```
zoobot, line 74: Does the address search show the distance of the searched for address to the current url's coords? Is that what this is for?

client/src/components/Header/Header.js
```diff
@@ -19,11 +24,8 @@ const Header = () => {
 
   return (
     <div id="header" className="header">
+      {/* Header elements are shown in row-reverse order, right to left */}
```
zoobot, line 27: How come it reversed?

client/src/components/Header/Header.js
```diff
@@ -77,6 +79,12 @@ const Header = () => {
             <AuthButton />
           </MenuItem>
         </Menu>
+
```
zoobot, line 82: <Search /> going in here somewhere?

client/src/pages/Search/Search.js
```diff
@@ -0,0 +1,88 @@
...
+
+const Search = ({ map }) => {
+  const [query, setQuery] = useState('');
+  const {
+    data: searchResults,
+    isLoading,
+    isError,
+    isSuccess,
+  } = useQuery({
+    queryKey: query,
```
zoobot, line 22: nice!! Glad you are bringing this back in here.

zoobot (approved):
Looks good, nice work!!

I left comments in this PR and #573 , feel free to merge separately.

client/src/pages/Search/Search.js
mwpark2014: Haven't ran into any problems - if you have access to a mapbox api dashboard, could you take a look to make sure everything is okay and not incurring charges? I probably used up like ~100 requests for testing, and requests under 100,000 should all be free

client/src/pages/Search/Search.js
mwpark2014: This biases the search towards results close to where the map is currently centered. It's actually pretty cool and I plan to make this smarter in future work. 

Tldr; this proximity makes address results so much more accurate when I'm looking at somewhere on the map that is close by my search query. It actually makes it less accurate when searching after initial zoomed out load since it searches for results around middle America

client/src/pages/Search/Search.js
zoobot: stats look fine. Let me know if you ever need a token or access to mapbox, we can make one for testing.

client/src/pages/Search/Search.js
zoobot: Interesting and cool!!

client/src/components/Header/Header.js
mwpark2014: This is to take out the `WATER THE TREES` home page link on the map for mobile since the searchbar will be overlayed on top of it for now. It solves the problem of positioning the hamburger menu icon on the right without using `float: right`. Simply remove the other elements in the flexbox row.

client/src/components/Header/Header.js
mwpark2014: It's not in the `Header` component. It's being shoved up there with absolution css positioning 😅 . Reason is that we don't want the map SearchBar to be on any page besides the initial map page. However, the header is currently the same for all pages. I also think we'll be removing the green header at some point, so this seemed like a fine solution to me for now.

mwpark2014: Yeah, let's make the `SearchBar` be a truly generic presentational component. Each page should be able to display the it - will just need to have a separate component to manage how to generate the search results and feed it into `SearchBar`, since that behavior differs between data, sources, and map

PR #577, opened 2023-03-20
Fix broken jest tests due to mismatch in dependency trees between Github CI and local

mwpark2014:
- Commit lock files to source control

These lock files shouldn't belong in source control.
https://docs.npmjs.com/cli/v6/configuring-npm/package-locks

https://github.com/waterthetrees/wtt_front/actions

FYI: When updating or creating new tests, you will likely need to pull these changes and `rm -rf node_modules; npm install;`. Otherwise, these mismatches might occur again and fail tests ran anywhere outside your local env


zoobot (approved):
Nice find!

PR #583, opened 2023-03-24
Fixed dev env setup so it points to dev server again

zoobot:


package.json
```diff
@@ -99,10 +99,10 @@
     "build:dev": "npm run build:data && webpack --env development",
     "docker": "docker-compose up",
     "watch": "webpack --watch --env development",
-    "start": "node server/index.js localserver",
+    "start": "node server/index.js localhost",
```
mwpark2014, line 102: oooh what changed here?

mwpark2014 (approved):
Overall lgtm if it was tested!

package.json
zoobot: just a rename to localhost really because localserver is a made up thing... and it seemed more of a pain in the neck to remember.
I'd already changed it in the server so this is bug fixing the package script.

PR #591, opened 2023-03-28
Minor css fixes for header and search for mobile tiny screens

zoobot:


client/src/pages/Search/Search.scss
```diff
@@ -1,16 +1,20 @@
...
-@media (max-width: 768px) {
-  .search-container {
+  @media (max-width: 768px) {
     // Screen width - (menu button width + padding)
     width: calc(100% - 70px);
     margin: 0;
     left: 8px;
   }
+
+  @media (min-width: 0em) and (max-width: 50em) {
```
mwpark2014, line 15: Seems like a non-standard screen separation. Super tiny screen? I'm hesitant to allow this go through. If we want to fix the current bug, that's fine, but I wouldn't want to open the floodgates for people defining all sorts of arbitrary media query breakpoints

https://www.freecodecamp.org/news/css-media-queries-breakpoints-media-types-standard-resolutions-and-more/ recommends the below, which I think is fairly standard:
320px — 480px: Mobile devices
481px — 768px: iPads, Tablets
769px — 1024px: Small screens, laptops
1025px — 1200px: Desktops, large screens
1201px and more —  Extra large screens, TV

zoobot: ok changed it to standard mobile, thanks

mwpark2014 (approved):
🚢

Issue #266, opened 2022-07-14
TREE - tree selfies


zoobot:

zoobot:
Hi @Zhikky sounds great if you want to work on this!! 
Currently there's images coming in here: https://github.com/waterthetrees/wtt_front/blob/main/client/src/pages/Tree/TreeImage.js They are just coming from wikimedia, not actual images of the street tree. 

We have not talked much about how this feature should work yet. 
Feel free to come to our next meeting and discuss how to get it to work. If you already have given it thought and have a good idea of how it should work, feel free to give it a go. 

Ideas:
It would be nice to have a "take picture of tree" button that then just loads the image from phone right away. 
Possibly we could just store all the images on the front end and then save the url to the db. We could potentially add a url field to history or something.

ri0nardo:
@Zhikky Wanted to know if there was any progress with this issue and if you had any additional questions.

ri0nardo:
@Zhikky hows progress on the tree selfies? Any questions?

PGrad:
@zoobot The MVP of the tree upload feature would be like this:
### Uploading an image
On the frontend, anyone can upload an image on any tree as long as they are signed in. The tree page will have an upload button that the user can click to upload an image, and this button will only appear when the user is signed in. On upload, a thumbnail is created along with the image, and the image and thumbnail are converted into uris and sent to the db. Whenever someone uploads an image, that image upload is added to the tree history and the history will link to the img.

Luke and I initially thought that only users who adopted a tree should be able to upload images for them, but the WTT team argued that some users may simply want to upload images in their neighborhood without the additional responsibility of adopting trees. Luke and I also thought about only allowing uploads for certain categories of trees - not allowing uploads for dead or stump trees for example - but there may be use cases for uploading images of dead or stump trees too.

Luke will handle the backend side of this, and they may create a separate write server which just handles image uploads, so the server which is handling reads doesn't get overloaded.

### Viewing images
A tree's page will show all of its images in a carousel. The images could show the date uploaded so users have a clear idea of the timeline of the tree.

### Nice to Haves
We could add a social media share feature, so users could take tree selfies and share them.

PGrad:
[...]

PR #592, opened 2023-03-29
Issue #266 Part 1: Update React 17->18

PGrad:
### Updated to React 18
```
package.json
package-lock.json
```
I see that Material UI doesn't have an off-the-shelf Carousel component to use, so I've been looking around online for components to use. I found [`react-spring-carousel`](https://react-spring-carousel.emilianobucci.com/docs) and I like it because it offers a [Headless UI](https://headlessui.com/) approach to the Carousel. That is to say, it doesn't come already styled and then you need to override those styles and do `!important` everywhere.

The problem is that `react-spring-carousel` needs React 18, so I had to update `react` and `react-dom` [per the instructions](https://react.dev/blog/2022/03/08/react-18-upgrade-guide). I also had to remove `@mui/styles` because it's been [deprecated in Material UI v5](https://mui.com/material-ui/migration/v5-style-changes/) and doesn't work with React 18. We're already using Material UI v5 and AFAIK we don't need the `@mui/styles library` anyway.

### Updated Jest test snapshots and updated npm packages
```
package.json
package-lock.json

client/src/tests/__snapshots__/
    - Header.test.js.snap
    - Searchbar.test.js.snap
    - Tree.test.js.snap
```
I ran `npm run start:dev` with `npm run watch` to run the local site, and everything works fine. I did `npm run test` and saw that there were some outdated snapshots and updated them.

I saw when pushing commits that Dependabot had also found some security vulnerabilities in outdated npm packages, so I updated them with `npm update`. Code janitor work.


zoobot (approved):
LGTM, thanks @PGrad

PR #594, opened 2023-03-29
Zoobot/feature/treefilter

zoobot:


ri0nardo:
please see the screenshot for my comments. These are all recommendations, these aren't necessary for the MVP.

PGrad (approved):
I did some testing of the two views, tested the filters, clicked on cards to see the details, tried downloaded the csv files and tested how responsive the page was to changes in screen size. It seems like the card view can be a little slow to update when changing the screen size, but overall the tree filter page works quite well.

PR #595, opened 2023-03-30
Issue #266: fix ReactDOM.render error

PGrad:
### Use createRoot() instead of ReactDOM.render()
React 18 uses createRoot() now instead of ReactDOM.render(). I make the replacement to fix the error.

### Use `font-display: swap` to prevent invisible text
The Lighthouse report shows the error `Ensure text remains visible during webfont load`. The problem is that if you have text being rendered in a certain font, it might be invisible for a moment because the font hasn´t loaded. The fix is to use `font-display: swap` so that the text will be rendered with a system font until the correct font has loaded.

### DIFF
--- a/client/src/index.js
+++ b/client/src/index.js
@@ -1,7 +1,8 @@
 import React from 'react';
-import ReactDOM from 'react-dom';
+import { createRoot } from 'react-dom/client';
 // The webpack.resolve.alias setting in webpack.config.js aliases '@' to 'client/src', so that all
 // the import paths can start from there, instead of being relative to the current file.
 import App from '@/components/App/App';
 
-ReactDOM.render(<App />, document.querySelector('.root'));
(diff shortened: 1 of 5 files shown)


client/src/index.js
zoobot, line 1: I checked out this branch on https://dev.waterthetrees.com and seeing the loading page pop up. I don't think it ws doing that before React 18, not sure if its actually loading slower or just using Suspense differently or what. 
If you get a chance can you do some performance/lighthouse tests between the two? I will as well and we can compare.

PGrad: @zoobot in the meantime, could you please merge this PR?

zoobot (approved):

client/src/index.js
zoobot: Thanks for your quick work on this!!

PR #596, opened 2023-03-31
test: issue 438: Bring accessibility up to 100% in Lighthouse check

PGrad:
### Abstract
There are two parts to this: fixing the outstanding accessibility issues that Lighthouse reports, and then adding Lighthouse to our CI to ensure that there are no regressions to accessibility in the future.

### Fixing accessibility issues
```
client/src/
    - index.html
    - components/Header/Header.js
    - tests/__snapshots__/Header.test.js.snap
    
    pages/
        - NewTree/NewTreeButton.js
        - UserLocation/GeolocateControl.js
```
There were two types of errors. The first one has to do with some buttons not having [a name which can be seen by screen readers](https://dequeuniversity.com/rules/axe/4.4/button-name). There are multiple ways to fix this, and my solution was to add a `title` attribute to the buttons. The side effect of this is that these buttons now also show tooltip text matching the title. @ri0nardo for reference, these are the buttons changed:

The Menu button:
[Screen recording 2023-03-30 16.55.30.webm](https://user-images.githubusercontent.com/117248915/228990099-caf90d76-8518-4263-b23b-b73b1283f25f.webm)

the Geolocate button:
[Screen recording 2023-03-30 16.54.58.webm](https://user-images.githubusercontent.com/117248915/228989963-4c3aa426-1381-4c0e-879c-bf715d47e05d.webm)

and the `NewTreeButton`:
[Screen recording 2023-03-30 16.56.13.webm](https://user-images.githubusercontent.com/117248915/228990551-ef67ec93-be1d-4a6e-83e1-7f7f42061dcf.webm)

Adding the title required updating the `Header.test.js` snapshot.

The other type of error was that [zoom should be enabled on the map](https://screpy.com/user-scalable-no-and-maximum-scale-errors/).

## Adding Lighthouse to CI
```
lighthouserc.js
.github/workflows/lighthouse-ci.yaml
server/index.js
```

It's great to be able to fix these issues, but the only way you can make the fix last is to add regression testing. I accomplish this by adding Lighthouse to our CI pre-flight. The Lighthouse test will fail if the scores for accessibility or best practices are under 100, so future developers submitting PRs will need to make sure that their UI features meet the standards of Lighthouse. The best, recommended practice for Lighthouse CI is to require 100 from the SEO and PWA scores as well, as performance can give variable results, but fixing those issues is a whole separate issue of work.

### DIFF
--- a/.github/workflows/lighthouse-ci.yaml
+++ b/.github/workflows/lighthouse-ci.yaml
@@ -1,0 +1,20 @@
+name: Build project and run Lighthouse CI
+on: [push]
+jobs:
+  lhci:
+    name: Lighthouse CI
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js 18.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 18.x
+      - name: npm install production
+        run: |
+          npm install --production
+      - name: run Lighthouse CI
+        run: |
+          npm install -g @lhci/cli@10.1.0
+          npm run build
+          lhci autorun --upload.target=temporary-public-storage || echo "LHCI failed!"
--- a/.github/workflows/lighthouse-ci.yaml
+++ b/.github/workflows/lighthouse-ci.yaml
@@ -4,15 +4,6 @@ jobs:
   lhci:
     name: Lighthouse CI
     runs-on: ubuntu-latest
-    container:
-      image: browserless/chrome
-      options: --user root
-      env:
-        NODE_ENV: development
-      ports:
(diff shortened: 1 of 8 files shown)


PGrad:
> Where do you want the token?

Add it as a Github env variable with the name `LHCI_GITHUB_APP_TOKEN`. See the Lighthouse docs [here](https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/getting-started.md#github-app-method-recommended). You can also email the token to me and I'll do it [paulgrad3rd@gmail.com](mailto:paulgrad3rd@gmail.com).

PR #598, opened 2023-03-31
Add DonateBtn component with basic styling and Open Collective link

Tijani-zainab:
mobile
desktop

### DIFF
--- a/client/src/pages/Map/DonateBtn/DonateBtn.js
+++ b/client/src/pages/Map/DonateBtn/DonateBtn.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import './DonateBtn.scss';
+
+const DonateBtn = () => {
+  return (
+    <div className="donate-btn">
+      <a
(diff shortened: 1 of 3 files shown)


zoobot (approved):
Nice work @Tijani-zainab 
Cool that we didn't have to use that giant button. 

@ri0nardo Any changes UI/design wise?

ri0nardo:
I would change it to be a white button with black text. Green is too primary. Also the white outline is somewhat odd to me.

ri0nardo:
> Something like this for mobile? https://www.pngitem.com/middle/ibmmxoT_donate-icon-pink-hd-png-download/

I was thinking like this. 
donate

ri0nardo:
@Tijani-zainab im testing the donate button on the dev site. When i hover it turns to white which makes it so you cant read the text. If we do a hover color change can we make it so the color white goes to a gray?

Tijani-zainab:
Ohh, yes I am. will make the changes

PR #606, opened 2023-04-02
test: Bring accessibility up to 100% in Lighthouse check

### LINKED ISSUE #438: Bring accessibility up to 100% in Lighthouse check
Bring accessibility up to 100% in Lighthouse check

Geolocate and plant buttons need `type="button"`

Test if removing this causes map problems: [user-scalable="no"]

PGrad:
Fixes #438.

### Abstract
There are two parts to this: fixing the outstanding accessibility issues that Lighthouse reports, and then adding Lighthouse to our CI to ensure that there are no regressions to accessibility in the future.

### Fixing accessibility issues
```
client/src/
    - index.html
    - components/Header/Header.js
    - tests/__snapshots__/Header.test.js.snap
    
    pages/
        - NewTree/NewTreeButton.js
        - UserLocation/GeolocateControl.js
```
There were two types of errors. The first one has to do with some buttons not having [a name which can be seen by screen readers](https://dequeuniversity.com/rules/axe/4.4/button-name). There are multiple ways to fix this, and my solution was to add a `title` attribute to the buttons. The side effect of this is that these buttons now also show tooltip text matching the title. @ri0nardo for reference, these are the buttons changed:

The Menu button:
[Screen recording 2023-03-30 16.55.30.webm](https://user-images.githubusercontent.com/117248915/228990099-caf90d76-8518-4263-b23b-b73b1283f25f.webm)

the Geolocate button:
[Screen recording 2023-03-30 16.54.58.webm](https://user-images.githubusercontent.com/117248915/228989963-4c3aa426-1381-4c0e-879c-bf715d47e05d.webm)

and the `NewTreeButton`:
[Screen recording 2023-03-30 16.56.13.webm](https://user-images.githubusercontent.com/117248915/228990551-ef67ec93-be1d-4a6e-83e1-7f7f42061dcf.webm)

Adding the title required updating the `Header.test.js` snapshot.

The other type of error was that [zoom should be enabled on the map](https://screpy.com/user-scalable-no-and-maximum-scale-errors/).

## Adding Lighthouse to CI
```
lighthouserc.js
.github/workflows/lighthouse-ci.yaml
server/index.js
```

It's great to be able to fix these issues, but the only way you can make the fix last is to add regression testing. I accomplish this by adding Lighthouse to our CI pre-flight. The Lighthouse test will fail if the scores for accessibility or best practices are under 100, so future developers submitting PRs will need to make sure that their UI features meet the standards of Lighthouse. The best, recommended practice for Lighthouse CI is to require 100 from the SEO and PWA scores as well, as performance can give variable results, but fixing those issues is a whole separate issue of work.

To see the results of our Lighthouse runs in pre-flight, I installed the [Lighthouse Github app](https://github.com/apps/lighthouse-ci) and run the action with the [`LHCI_GITHUB_APP_TOKEN`](https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/getting-started.md#github-app-method-recommended) env variable. You should be seeing a status in the pre-flight with the Lighthouse scores and the icon below.

Actually getting the github action to work was quite difficult, mainly because Lighthouse needs to have chrome installed and the different solutions I tried to install Chrome didn't seem to work (using a [docker image](https://hub.docker.com/r/browserless/chrome), using a [github action](https://github.com/treosh/lighthouse-ci-action)) etc. In the end, just manually downloading the chrome .deb debian file worked best, although I fear pulling the .deb by wget could be flaky if it 404s.

I used [nektos/act](https://github.com/nektos/act) to test my github action locally, so I could get faster turn-around and _not_ submit a hundred github commits just to test my github action. That approach has its limits too though, as the action would fail due to a PROTOCOL_TIMEOUT error that is apparently the [#1 most common issue](https://github.com/GoogleChrome/lighthouse/issues/6512) with Lighthouse.

Overall, I like that we can do automated testing with Lighthouse CI, but the tool seems a lot buggier than I would expect from Google devs. It's good enough.

### DIFF
--- a/.github/workflows/lighthouse-ci.yaml
+++ b/.github/workflows/lighthouse-ci.yaml
@@ -1,0 +1,20 @@
+name: Build project and run Lighthouse CI
+on: [push]
+jobs:
+  lhci:
+    name: Lighthouse CI
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js 18.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 18.x
+      - name: npm install production
+        run: |
+          npm install --production
+      - name: run Lighthouse CI
+        run: |
+          npm install -g @lhci/cli@10.1.0
+          npm run build
+          lhci autorun --upload.target=temporary-public-storage || echo "LHCI failed!"
--- a/.github/workflows/lighthouse-ci.yaml
+++ b/.github/workflows/lighthouse-ci.yaml
@@ -5,9 +5,9 @@ jobs:
     name: Lighthouse CI
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
       - name: Use Node.js 18.x
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
         with:
           node-version: 18.x
       - name: npm ci
--- a/.github/workflows/lighthouse-ci.yaml
+++ b/.github/workflows/lighthouse-ci.yaml
@@ -4,15 +4,6 @@ jobs:
   lhci:
     name: Lighthouse CI
     runs-on: ubuntu-latest
-    container:
-      image: browserless/chrome
-      options: --user root
-      env:
-        NODE_ENV: development
-      ports:
-        - 80
-      volumes:
-        - /docker_volume
     steps:
       - uses: actions/checkout@v1
       - name: Use Node.js 18.x
@@ -21,6 +12,11 @@ jobs:
           node-version: 18.x
       - name: npm ci
         run: npm ci
+      - name: Install chrome
+        run: |
+          wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
+          sudo apt-get update
+          sudo apt-get install -y ./google-chrome-stable_current_amd64.deb
       - name: Run Lighthouse CI
         run: |
           npm install -g @lhci/cli@0.11.x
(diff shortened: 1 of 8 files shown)


.github/workflows/lighthouse-ci.yaml
```diff
@@ -0,0 +1,27 @@
...
+          node-version: 18.x
+      - name: npm ci
+        run: |
+          npm ci
+          npm run build
+      - name: Install chrome
+        run: |
+          wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
+          sudo apt-get update
+          sudo apt-get install -y ./google-chrome-stable_current_amd64.deb
```
zoobot, line 21: How long does this wget, update and install take to run every time?

lighthouserc.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+  ci: {
+    collect: {
+      startServerCommand: 'npm run start:dev',
+      url: ['http://localhost:3001'],
```
zoobot, line 5: webpack serve is on 3000 so this will need a change once your reloading is fixed.

.github/workflows/lighthouse-ci.yaml
PGrad: The `Install chrome` step takes about 33 seconds on CI, and the whole Github action for Lighthouse CI takes about 4 and a half minutes.

zoobot (approved):
How long does this take?

.github/workflows/lighthouse-ci.yaml
zoobot: Takes about 1m 42s

PR #622, opened 2023-04-05
Add ability to search for lat/long in map search bar

mwpark2014:
Add ability to search for lat/long coordinates in the map search bar!
Doesn't incur mapbox request (unless the person types really slow and requests are sent before the code realizes it's a lat/long query)

image
This is how Google does it

image
This is how we do it

### DIFF
--- a/client/src/pages/Search/Search.js
+++ b/client/src/pages/Search/Search.js
@@ -135,7 +135,7 @@ const isValidFloat = (token) => {
 // Use the first delimiter in VALID_LAT_LNG_DELIMITERS that is found in the input query string
 // to split string into potential lat/lngs
 const splitWithValidDelimiter = (query) => {
-  let coords = query;
+  let coords = [query];
   VALID_LAT_LNG_DELIMITERS.forEach((delimiter) => {
     if (query.includes(delimiter)) {
       coords = query.split(delimiter);
--- a/client/src/pages/Search/Search.js
+++ b/client/src/pages/Search/Search.js
@@ -10,6 +10,7 @@ const MIN_CHARS_FOR_SEARCH = 2;
 const BASE_MAPBOX_SEARCH_API_URL =
   'https://api.mapbox.com/geocoding/v5/mapbox.places/';
 const SEARCH_QUERY_CACHE_TIME = 1000 * 30; // 30 seconds
+const VALID_LAT_LNG_DELIMITERS = [',', '/', '\\'];
 
 const Search = ({ map }) => {
   const [query, setQuery] = useState('');
@@ -99,8 +100,7 @@ const getMapboxSearchResults = async (query) => {
 
 // Attempt to parse a latitude and longitude from the query
 export const isQueryLatLong = (query) => {
-  // Use commas and forward slashes as possible delimiters
-  const coords = query.includes(',') ? query.split(',') : query.split('/');
+  const coords = splitWithValidDelimiter(query);
   if (coords.length !== 2) {
     return false;
   }
@@ -132,9 +132,22 @@ const isValidFloat = (token) => {
   return true;
 };
 
+// Use the first delimiter in VALID_LAT_LNG_DELIMITERS that is found in the input query string
+// to split string into potential lat/lngs
+const splitWithValidDelimiter = (query) => {
+  let coords = query;
+  VALID_LAT_LNG_DELIMITERS.forEach((delimiter) => {
+    if (query.includes(delimiter)) {
+      coords = query.split(delimiter);
+      return;
+    }
+  });
+  return coords;
+};
+
 // Return a SearchResult that is compatible with the feature data shape from mapbox search API
 const createLatLongSearchResult = (query) => {
-  const coords = query.includes(',') ? query.split(',') : query.split('/');
+  const coords = splitWithValidDelimiter(query);
   const latLng = coords.map((coord) => parseFloat(coord));
 
   return {
--- a/client/src/pages/Search/Search.js
+++ b/client/src/pages/Search/Search.js
@@ -37,6 +37,7 @@ const Search = ({ map }) => {
     debouncedSetQuery(newQuery);
   };
 
+  // FIXME: This doesn't get triggered when the current selection is selected again from the autocomplete.
   const handleOptionSelect = (e, option) => {
     if (option) {
       map.panTo(option.coords);
--- a/client/src/pages/Search/Search.js
+++ b/client/src/pages/Search/Search.js
@@ -23,14 +23,13 @@ const Search = ({ map }) => {
   } = useQuery({
     queryKey: query,
     queryFn: () => getMapboxSearchResults(query),
-    enabled: query.length > MIN_CHARS_FOR_SEARCH,
     cacheTime: SEARCH_QUERY_CACHE_TIME,
   });
 
   // Debounce search requests to mitigate churning through our API requests budget
   const debouncedSetQuery = useMemo(
-    (setQuery) => getDebouncedSetQuery(setQuery),
-    [],
+    () => getDebouncedSetQuery(setQuery),
+    [setQuery],
   );
   const handleInputChange = (event) => {
     const newQuery = event.currentTarget.value;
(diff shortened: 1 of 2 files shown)


client/src/pages/Search/Search.js
```diff
@@ -85,4 +86,62 @@ const getSearchResults = async (query) => {
...
+    address: result.place_name,
+    id: result.id,
+    coords: result.center,
+    type: 'Results',
+  }));
+};
+
+// Attempt to parse a latitude and longitude from the query
+export const isQueryLatLong = (query) => {
+  const tokens = query.split(',');
```
zoobot, line 102: Can we do an includes , or / here so users can copy lng/lat from the browser into search?

zoobot, line 102: Can we call this variable coords or coordinates instead of tokens to be more clear at a glance what this is doing?

zoobot (changes requested):
This is great! LGTM trying this out. Requested minor changes.

zoobot (commented):
Nice tests!

client/src/pages/Search/Search.js
mwpark2014: yes, perfect

mwpark2014:
> This is great! LGTM trying this out. Requested minor changes.

This has been updated! Merge whenever ready

client/src/pages/Search/Search.js
```diff
@@ -85,4 +87,75 @@ const getSearchResults = async (query) => {
...
+  const longitude = parseFloat(coords[1]);
+  // Longitude must a number between -180 and 180
+  if (Math.abs(longitude) > 180) {
+    return false;
+  }
+
+  return true;
+};
+
+const isValidFloat = (token) => {
```
zoobot, line 127: This is a nice reusable function. In this code base context, our use cases most likely will be for coords so maybe could use coords as variable name here for consistency.

zoobot (dismissed):

zoobot (commented):
```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `ForwardRef(Autocomplete)`. See https://reactjs.org/link/warning-keys for more information.
    at SearchResult (webpack-internal:///./client/src/components/Search/SearchBar.js:89:22)
    at Autocomplete (webpack-internal:///./node_modules/@mui/material/Autocomplete/Autocomplete.js:400:83)
    at SearchBar (webpack-internal:///./client/src/components/Search/SearchBar.js:25:24)
    at div
    at Search (webpack-internal:///./client/src/pages/Search/Search.js:30:18)
    ```
    
    Can you add a key for Autocomplete or is this a MUI error?

zoobot (changes requested):
Getting this issue randomly.
Screen Shot 2023-04-06 at 8 46 22 AM

To replicate _maybe_?, start putting in one coordinate into the search, then a comma or slash, then go start to copy the other coordinate. It's happened twice when I was doing something like that..

mwpark2014:
> Getting this issue randomly. Screen Shot 2023-04-06 at 8 46 22 AM
> 
> To replicate _maybe_?, start putting in one coordinate into the search, then a comma or slash, then go start to copy the other coordinate. It's happened twice when I was doing something like that.. I think its probably hitting autocomplete or something before there's both coords in there

sigh, I know what's wrong. I would've hoped that it was caught in the tests. My bad. Thanks for catching.

mwpark2014:
This PR is ready to merge. I added the breaking test case (I tested by copy and pasting a lat/long and testing some normal string queries, but I missed the test case of typing in a number. It failed since strings can be iterated through but not numbers in javascript...)

mwpark2014:
> ```
> Warning: Each child in a list should have a unique "key" prop.
> 
> Check the render method of `ForwardRef(Autocomplete)`. See https://reactjs.org/link/warning-keys for more information.
>     at SearchResult (webpack-internal:///./client/src/components/Search/SearchBar.js:89:22)
>     at Autocomplete (webpack-internal:///./node_modules/@mui/material/Autocomplete/Autocomplete.js:400:83)
>     at SearchBar (webpack-internal:///./client/src/components/Search/SearchBar.js:25:24)
>     at div
>     at Search (webpack-internal:///./client/src/pages/Search/Search.js:30:18)
> ```
> 
> Can you add a key for Autocomplete or is this a MUI error?

So how big of a priority is this to you? This is a non-breaking warning, and this is coming from how the MUI autocomplete works. For some reason, it is stripping the `key` that is defined in the code already for `SearchResult` (you can see it defined in the code). I have a hunch that `renderOption` and `renderGroup` are not supposed to be used together (MUI documentation is lacking...), but I've already spent some time on this and it probably isn't worth the effort to fix IMO

zoobot (changes requested):
Small change request I hope? 
This doesn't remember lng/lat after editing. Remembering the input is expected behavior. 

When searching google maps by lng/lat it remembers the lnglat put in, and if edited, it changes the input to match the new lnglat

https://www.google.com/maps/place/37%C2%B046'28.6%22N+122%C2%B017'56.4%22W/@37.7746041,-122.301194,17z/data=!3m1!4b1!4m4!3m3!8m2!3d37.7746041!4d-122.299

On ours when a lntlat is entered to the input search, say ```37.7746041,-122.301194``` and then edit the input to ```37.7746041,-122.44``` it will change the input to an address, I think the top address that it's found in address search? It should remain a lng/lat in the input.

mwpark2014:
> Small change request I hope? This doesn't remember lng/lat after editing. Remembering the input is expected behavior.
> 
> When searching google maps by lng/lat it remembers the lnglat put in, and if edited, it changes the input to match the new lnglat
> 
> https://www.google.com/maps/place/37%C2%B046'28.6%22N+122%C2%B017'56.4%22W/@37.7746041,-122.301194,17z/data=!3m1!4b1!4m4!3m3!8m2!3d37.7746041!4d-122.299
> 
> On ours when a lntlat is entered to the input search, say `37.7746041,-122.301194` and then edit the input to `37.7746041,-122.44` it will change the input to an address, I think the top address that it's found in address search? It should remain a lng/lat in the input.

I think I understand, but I want to get a bit more clarity for whenever I dig into this issue: 
Are you typing the lat/long manually or copy and pasting? And when editing, is the longitude completely deleted before editing or are you just editing the digits after the decimal?

mwpark2014:
> copy and pasting, then editing the digits after the decimal. Might be an edge case as I am not sure how many people actually do this. I do it often enough but I'm probably not the norm on this one.

Okay, I can't repro those exact conditions. Copy and pasting a lat/long and editing should be supported in the way you expect. I'm going to infer that you typed out a query that could be perceived as an address at some point. When we make a lat/lng query, it doesn't clear out the mapbox search results. Should be an easy fix

> This doesn't remember lng/lat after editing. Remembering the input is expected behavior.

This doesn't make sense to me. I think the below screenshot is what you're referring to. Can you confirm? In this case, the lat/lng is remembered (it's just the last result). 
image

mwpark2014:
Updated:

https://user-images.githubusercontent.com/10445556/230791917-49497619-6d01-4187-a2e8-3c09969051e9.mp4

Please let me know if there's anything else! Any minor non-blockers can go into a future PR :)

mwpark2014:
> > > copy and pasting, then editing the digits after the decimal. Might be an edge case as I am not sure how many people actually do this. I do it often enough but I'm probably not the norm on this one.
> > 
> > 
> > Okay, I can't repro those exact conditions. Copy and pasting a lat/long and editing should be supported in the way you expect. I'm going to infer that you typed out a query that could be perceived as an address at some point.
> 
> Yeah I think you are right that it is thinking the lat/long's a street address and so it puts that street address in instead. I will try to reproduce and video it. Adding it to a later PR sounds just fine, its an edge case use.

Yeah, this should be fixed along with the bug that was merged in with the other fix lol

mwpark2014:
I think all in all, I'm trying to minimize the amount of time I spend on incorporating small improvements into these PRs. If we could batch all the small asks together or be okay with them going in future PRs, I think relying on manual testing + unit testing is reasonable. Otherwise, I'll write an integration test in the future since it takes quite a while to repeatedly manually test through all the test cases for every small change, esp with the slow local dev webpack reloading at the current moment. 

Anyway, the PR has been tested manually and should be ready to be merged now @zoobot

PR #637, opened 2023-04-08
Refactor styles:

Tijani-zainab:
Consolidate .donate-btn class and remove unnecessary style


client/src/components/Button/DonateBtn/DonateBtn.scss
```diff
@@ -0,0 +1,15 @@
+.donate-btn {
+  text-align: center;
+  margin: 5px;
+}
+
+a {
```
zoobot, line 6: One more tiny baby change... 
Can you add  text-decoration: none; for a:hover?


Here's an article about sass nesting to control scope if you want to go that route
https://css-tricks.com/using-sass-control-scope-bem-naming/

Used like this with an a:
```
.donate-btn {
  text-align: center;

  & a {
    color: black;
    text-decoration: none;
  }
  & a:hover {
    color: darkgreen; //the text-decoration is inherited from above with this nested.
  }
}

Tijani-zainab: Updated it, I should say, this SASS syntax is way better! Honestly, I should've used it right from the get-go!

zoobot: :) Me too! I learned it from Vern. Thanks for making the update.

zoobot: Can you add a test for this button? Here's an example of the other button test:
https://github.com/waterthetrees/wtt_front/blob/main/client/src/tests/WttButton.test.js

to run you just type `npm run test` and if you just want to run your test you can `npm run test yourtest`

Tijani-zainab: > Can you add a test for this button? Here's an example of the other button test: https://github.com/waterthetrees/wtt_front/blob/main/client/src/tests/WttButton.test.js
> 
> To run you just type `npm run test` and if you just want to run your test you can `npm run test yourtest`. Don't worry about writing a test for the map part of it as that is a much larger task. You should be able to just test the Donate component button on its own.

Ohh, I totally missed this. I apologize! 
I almost send a message yesterday regarding why we are still holding off on merging it.

Tijani-zainab: I'm really Sorry, I'll get to it tonight!

Tijani-zainab: for two weeks 😳😐🤦🏾‍♀️

zoobot: No worries!! No rush on this really. :)

Tijani-zainab: 🙌🙌

Tijani-zainab:
Thank You @zoobot for your your help!

zoobot (approved):
LGTM! Thanks for putting in the extra 10% to write the test!! :)

PR #648, opened 2023-04-14
Zoobot/feature/treelist showhide button

zoobot:
Treelist Epic:

Shows and hides the Treelist header on mobile so there's more room to see the tree card or list.


Tijani-zainab:
@zoobot Lgtm! I'm yet to test the branch out locally. but will do so now. 
Nice work!

client/src/tests/ButtonShowHide.test.js
```diff
@@ -0,0 +1,63 @@
...
+    const { getByAltText } = render(
+      <ButtonShowHide showMore={false} setShowMore={mockSetShowMore} />,
+    );
+    const downArrowBtn = getByAltText('show');
+    fireEvent.click(downArrowBtn);
+    expect(mockSetShowMore).toHaveBeenCalledWith(true);
+  });
+
+  it('Handles up arrow (hide) button click', () => {
+    const { getByAltText } = render(
```
mwpark2014, line 38: Nice, I need to write these kinds of tests

mwpark2014 (commented):
It might be helpful to add a description on the problem these PRs are trying to solve. I'm not sure where this component is going to be used, so I'm not sure if this feedback applies:

I think the one thing I'm concerned about is the expanding library of custom components that might already have implementations in mui. Have you considered using the `Collapse` component (https://codesandbox.io/s/8gv7jz?file=/demo.tsx:1693-1701) or something similar?

mwpark2014 (approved):
> I did try using Collapse but the custom css I'd need to write over the default seemed like more trouble than it was worth.
I'll a description and take a screenshot.

Okay, I think I can be okay with approving the PR to unblock these changes. I do have some concern that developers will build upon this custom component in the future, which will require updating it to be more extensible and add accessibility, when there might be a reusable component out there that does what we want, but I don't see an easy alternative in MUI and it's prob not worth the time right now

PR #650, opened 2023-04-14
Zoobot/feature/treelist card

zoobot:


client/src/components/Card/Card.scss
```diff
@@ -0,0 +1,29 @@
...
+  transition: box-shadow 0.3s ease-in-out;
+  margin-bottom: 1rem;
+  padding: 0 0 4px;
+  height: 100%;
+  background-color: whitesmoke;
+  color: gray;
+  break-inside: avoid;
+  border-radius: 4px;
+
+  h1,
```
mwpark2014, line 12: Could this and line 18-22 be combined? Also, are you using prettier here? My prettier styles this file differently (kinda weird that h1 is on a different line from the rest of the h[x]. I wonder if we want to add pretter and linting pre-commit checks to the github CI. 

These are all nits

mwpark2014 (approved):

client/src/components/Card/Card.scss
zoobot: I thought we do have prettierrc.js setup and then eslintrc uses it automatically... ?? When I run in manually it does format it differently though so I am not sure about this. I didn't think we had to run it manually.

zoobot: Can you post your .vscode setting file? Does yours do it automatically?

mwpark2014: I've added ` "editor.formatOnSave": true,` to my .vscode settings.json file. 

Added this issue https://github.com/waterthetrees/wtt_front/issues/654

zoobot: ran prettier and combined 18-22. Thanks for the review! @mwpark2014

PR #652, opened 2023-04-14
Zoobot/patch/treelist image downloader

zoobot:
Treelist Epic:

This is changing the current behavior of the build file to become forward compatible with the treelist.

Build normalizes the data json files and puts them in the dist folder.
It also calls wikipedia and creates a list of image urls for our tree side panel.

There is also added functions in this to download the wikipedia images for our treelist/data pages so we will have image cards and images for the new tree pages.

You need to run 
`npm run build `to run this build file.


client/src/pages/Tree/TreeImage.js
```diff
@@ -25,7 +27,20 @@ export const ImageLoad = React.memo(({ src, placeholder, alt = '' }) => {
       }}
       placeholder={placeholder}
       alt={alt}
+      loading="lazy"
     />
   );
 });
 ImageLoad.displayName = 'ImageLoad';
+
+export const fixScientificForImage = (scientific) => {
```
mwpark2014, line 36: Oh man, names like `scientific` makes me really want Typescript in our codebase. I have no idea what this parameter is supposed to represent upon first inspection. Had to search around to find that it's referring to scientific name. maybe `scientificName` or alternative is more informative?

client/src/pages/Tree/TreeImage.js
```diff
@@ -25,7 +27,20 @@ export const ImageLoad = React.memo(({ src, placeholder, alt = '' }) => {
...
+  if (!scientific || !treeImages[scientific]) return null;
+
+  // replace spaces with hyphens for image path using rejex
```
mwpark2014, line 39: nit: regex?

client/src/pages/Tree/TreeImage.js
mwpark2014: This method might need a comment describing why we would need to fix scientific for image

client/src/tests/TreeImage.test.js
```diff
@@ -0,0 +1,59 @@
...
+  const alt = 'Sample image';
+
+  it('Renders the ImageLoad component correctly', () => {
+    const { container } = render(
+      <ImageLoad src={src} placeholder={placeholder} alt={alt} />,
+    );
+    expect(container).toMatchSnapshot();
+  });
+
+  it('loads the actual image and updates the src', async () => {
```
mwpark2014, line 17: appreciate the attention given to these tests <3

client/src/tests/build.test.js
```diff
@@ -0,0 +1,42 @@
+// TODO - add tests for the build script
+// Currently this is just a placeholder file, we need to move the server to es6 modules and add tests
```
mwpark2014, line 2: Oh, this is interesting. Can you explain what you mean by 
> we need to move the server to es6 modules and add tests

data/build.js
mwpark2014, line 1: Could you give some context on when this file is called? It almost seems like something we should be running on the server on an infrequent cadence and providing a cached version to our front-end.

client/src/tests/build.test.js
zoobot: Sorry writing this quick as my friend's trying to get me to go running now.

We've moved wtt_server  and tree-sources over the es6 modules so to have same testing on this repo, we should do the same. Because the testing in React is using modules, to run backend tests, it'll make it easier. The reason we had to move is that the hashing of the tree id created totally different ids if we used one vs the other. Strange issue but can explain more when I get back.

data/build.js
zoobot: It gets called on npm run build and does exactly what you are saying, it builds a cached json files for us to use and downloads the wikipedia images for the tree side panel and soon the treelist(That PR is not yet up) I had to make changes to actually download the images so it wouldn't have to query wikipedia. It only runs the download on build, if it can't find the image, it'll query wikipedia. Also this build file normalizes the json files used for the treelist page(and currently the data page). We should separate those probably(the json creation and image download/creation) but that was a larger task I didnt want to take on immediately.

client/src/pages/Tree/TreeImage.js
zoobot: Added comment

client/src/pages/Tree/TreeImage.js
zoobot: fixed spelling

data/build.js
mwpark2014: Oh interesting. I'm surprised this isn't in `wtt_server`. The `/data` package in general seems like it should be provided by `wtt_server` since conceptually, that's where I would expect all data sources to live.

mwpark2014 (approved):
Comments aren't blocking

data/build.js
```diff
@@ -214,6 +232,50 @@ function replaceIrregularNames(name) {
...
+async function downloadImage(imageURL, title) {
+  try {
+    const response = await fetch(imageURL);
+    if (!response.ok) {
+      throw new Error(`Failed to download image: ${response.statusText}`);
+    }
+
+    const imageName = `${title.split(' ').join('-')}.jpg`;
+    const buffer = await response.arrayBuffer();
+    fs.writeFileSync(
```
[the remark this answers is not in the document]
zoobot: Ah nice catch! Thanks!

data/build.js
zoobot: Yeah, this sort of grew organically into what it is now. At first it was just a bunch of static csvs/json lists loaded via dropdown into the data page. Then John realized there were mushrooms and fish in the tree lists so he built some functions to get rid of those, then Jason added on with the wiki photos url grabber, then I added the photo downloader. We could store the json in the database. The lists have dynamic fields. All the stuff just loads as local static files. 

We should separate the json lists and the photos functionality. I could def see the photo names/urls being stored in taxonomy & images tables.

PR #653, opened 2023-04-15
Zoobot/feature/treelist license

zoobot:


mwpark2014:
FWIW, thanks for separating the PR into smaller chunks, but they don't have to be _this_ bite sized haha. Thanks for the effort though! 🚀

mwpark2014 (approved):

PR #657, opened 2023-04-16
Add new prettier-plugin for sorting imports

mwpark2014:
* Prettier-plugin will be a new dev dependency

Something that was triggering my OCD. This should introduce auto-sorting imports as part of our prettier setup. 
Sorts by 4 categories: 
* 3rd party deps
* Absolute paths starting with '@/'
* Relative paths starting with './',
* Relative paths with file suffixes such as '.css', '.scss', etc

image


PGrad (approved):
Import sorting? Trivago.

mwpark2014:
Improved on the regex (prob still not perfect) and added some comments to hopefully read it and understand it again one day

mwpark2014:
Ruh roh, the lighthouse CI is now failing for this PR, which certainly shouldn't be affected by this dev dependency

Issue #633, opened 2023-04-08
Zoom into tree on search


zoobot:
Zoom in when you do a search. 
Requested by FUF

mwpark2014:
@zoobot I added a new line item for expected search behavior in https://docs.google.com/document/d/1MSSbkCo77VCE9tHZdWKAr8lfrBpRF_1AEMptInFjUBI/edit#heading=h.m5b3bqzglb6q

The reason why zoom wasn't implemented was to stay consistent with other zoom behavior in similar actions such as geolocation. I propose that we follow the search behavior outlined in the document and we change the zoom behavior of other actions to match. I.e., passing in tree-id and activating geolocation both trigger a zoom-in.

zoobot:
LGTM, 
Once it's triggered, it should zoom initially but if a user zooms back out, it should remember the manual zoom.

zoobot:
Since this is a larger task, do you want to handle just the search zoom first or do you want to build out reusable zoom first?

mwpark2014:
> Once it's triggered, it should zoom initially but if a user zooms back out, it should remember the manual zoom.

What does that mean? What if someone backs out to the globe view again and then triggers another search to a particular address or lat/long? I think it should still zoom in in that situation. 

We need to apply a standard on when we should zoom or should not zoom due to user preferences. 

Scenarios: 
* We shouldn't zoom out if a user is completely zoomed (for example, zoomed in on a tree)
* If a user zooms out, they might not want to be zoomed in so much... but they might also want to keep that zooming functionality

Potential solutions:
* What if we try to be a bit smarter and record zoom preferences within specific thresholds? For example: "user typically zooms in to this level within 12-15. we can assume this is the zoom level they want if they activate geolocation or address search". 
* If a user attempts to cancel a zoom mid-zoom, we deactivate zooming for a specific type (types would be country-level, state, city, address/latlng, etc)
* Ignore complete personalization and just try to set sane zoom levels for each particular action. I.e., latlng search will always zoom, no matter what

Just spitting out potentially bad ideas

zoobot:
> > Once it's triggered, it should zoom initially but if a user zooms back out, it should remember the manual zoom.
> 
> What does that mean? What if someone backs out to the globe view again and then triggers another search to a particular address or lat/long? I think it should still zoom in in that situation.

[...]

PR #658, opened 2023-04-16
MapboxManager as a passthrough layer between components and MapBox API

mwpark2014:
Codifies the expectations set in [WTT Camera Actions Spec](https://docs.google.com/document/d/1MSSbkCo77VCE9tHZdWKAr8lfrBpRF_1AEMptInFjUBI/edit#)

This is a pretty significant change to how we access the Mapbox `map` dependency

### Summary
* New MapboxManager class that encapsulates access to the Mapbox `map` object, providing standard ways to get map info or manipulate the map
* This MapboxManager will be provided through a simple React Context (Intentionally avoiding burying the context and provider in a custom hook like we do with `useNewTree`). This will remove the tight coupling between several UI components and the map
* Holds reusable logic such as zooming behavior based on type of region

### Problem being solved:
* Defines a WTT-specific interface between our UI components and the Mapbox map object
* Standardize mapbox camera behavior and eventually get robust testing & monitoring in for user interactions on the map
* Presentational components should not be accessing the map themselves. It's less maintainable needing to identify every invocation of `map.flyTo` or `map.panTo`, etc when trying to track down bugs or make sweeping changes
* Testability is much easier after consolidation and reuse of WTT-specific map logic
 
### Makes it easy to add later on:
* Any common change across all the calls to `flyTo` or other mapbox API-related stuff
* Any common map optimizations
* Any common monitoring

Guidelines going forward:
* There are several components in `Map.js` that are very tightly coupled to the mapbox map such as `MapLayers`, `MapControlPortal`, `TreeLayerLegend`, and children to `MapLayers`. These are fine to pass `map` into since these are largely `MapBox` related concepts. 
* MapLayout should not receive `map` and should not pass down `map` to any of its children (siblings to `MapContainer`)
* All calls to mapbox camera actions (like `map.flyTo`, `map.panTo`, etc) should go through `MapboxManager`

Manually tested:
* Search (address, state, country, city, lat/lng)
* Activate geolocation
* Selecting trees, loading new page with selected tree id
* Planting new tree
* Clicking on tree count layer

Will hold on map integration tests since I was in the middle of working on that in another PR.
* Will test more cases as I come up with them through the week since this touches a lot of areas

Fixes https://github.com/waterthetrees/wtt_front/issues/636 and https://github.com/waterthetrees/wtt_front/issues/633


client/src/pages/Map/Map.js
```diff
@@ -81,6 +85,7 @@ export default function Map({
   onLoad,
 }) {
   const [map, setMap] = useState(null);
+  const mapboxManager = useContext(MapboxManagerContext);
   const [isMapLoaded, setIsMapLoaded] = useState(false);
```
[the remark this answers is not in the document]
mwpark2014: Hmm, but that's not exactly what I'm doing here lol

client/src/pages/Map/MapboxManagerProvider.js
```diff
@@ -0,0 +1,15 @@
+import React, { createContext } from 'react';
+
+import MapboxManager from './MapboxManager';
+
+export const MapboxManagerContext = createContext(undefined);
+
+const mapboxManager = new MapboxManager();
+
+const MapboxManagerProvider = ({ children }) => {
```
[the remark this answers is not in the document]
mwpark2014: Oh, this actually helped spot the issue I was trying to locate

mwpark2014:
@PGrad sweet, I love that we have the Lighthouse CI added. Is there a way to rerun these tests in case they're flaky?

PGrad:
> @PGrad sweet, I love that we have the Lighthouse CI added. Is there a way to rerun these tests in case they're flaky?

If you click on "Details" you can re-run the job; I just re-started the Lighthouse job.

mwpark2014:
Ready for review @zoobot. Feel free to suggest more manual test cases

mwpark2014:
> > @PGrad sweet, I love that we have the Lighthouse CI added. Is there a way to rerun these tests in case they're flaky?
> 
> If you click on "Details" you can re-run the job; I just re-started the Lighthouse job. Documentation: https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs
> 
> A simple "re-run" button would be nice though.
> 
> E: it looks like now your PR is not passing best practices because some errors are being logged to console, I think related to issue #660?

Now getting

```
[webpack-cli] TypeError: Cannot create proxy with a non-object as target or handler
    at new WrappedPlugin (/home/runner/work/wtt_front/wtt_front/node_modules/speed-measure-webpack-plugin/WrappedPlugin/index.js:280:12)
    at /home/runner/work/wtt_front/wtt_front/node_modules/speed-measure-webpack-plugin/index.js:[51](https://github.com/waterthetrees/wtt_front/actions/runs/4777182030/jobs/8492707526#step:4:52):14
    at Array.map (<anonymous>)
    at SpeedMeasurePlugin.wrap (/home/runner/work/wtt_front/wtt_front/node_modules/speed-measure-webpack-plugin/index.js:44:45)
    at module.exports (/home/runner/work/wtt_front/wtt_front/webpack.config.js:1[53](https://github.com/waterthetrees/wtt_front/actions/runs/4777182030/jobs/8492707526#step:4:54):30)
    at loadConfigByPath (/home/runner/work/wtt_front/wtt_front/node_modules/webpack-cli/lib/webpack-cli.js:1439:37)
    at async WebpackCLI.loadConfig (/home/runner/work/wtt_front/wtt_front/node_modules/webpack-cli/lib/webpack-cli.js:1510:38)
    at async WebpackCLI.createCompiler (/home/runner/work/wtt_front/wtt_front/node_modules/webpack-cli/lib/webpack-cli.js:1785:22)
    at async WebpackCLI.runWebpack (/home/runner/work/wtt_front/wtt_front/node_modules/webpack-cli/lib/webpack-cli.js:1890:20)
    at async Command.<anonymous> (/home/runner/work/wtt_front/wtt_front/node_modules/webpack-cli/lib/webpack-cli.js:912:21)
Error: Process completed with exit code 2.
```

Do you know what's wrong? I'm inclined to ignore for now and investigate at a later point

client/src/pages/Map/MapLayout.js
```diff
@@ -105,7 +105,9 @@ export default function WrappedMapLayout() {
     <ErrorBoundary>
       <NewTreeProvider>
         <UserLocationProvider>
-          <MapLayout />
+          <MapboxManagerProvider>
```
zoobot, line 108: At what point should we use redux. Should we be checking into combining contexts into a global useContext or do these providers not overlap much since they have very different use cases?

client/src/pages/Map/MapboxManager.js
```diff
@@ -0,0 +1,85 @@
...
+    if (!this.map) {
+      return;
+    }
+
+    return this.map.getCenter();
+  }
+
+  // @param {LngLatLike} coords: See https://docs.mapbox.com/mapbox-gl-js/api/geography/#lnglatlike
+  // (LngLat | {lng: number, lat: number} | {lon: number, lat: number} | [number, number])
+  setCenter({ coords, regionType }) {
```
zoobot, line 23: Where's regionType coming from?

client/src/pages/Map/MapboxManager.js
```diff
@@ -0,0 +1,85 @@
...
+        });
+        break;
+      default:
+        this.map.flyTo({
+          center: coords,
+        });
+    }
+  }
+
+  addMarkerRefToMap(markerRef) {
```
zoobot, line 77: What does this Marker refer to in our case?

client/src/pages/Map/TreeCountLayer.js
```diff
@@ -49,14 +52,12 @@ export default function TreeCountLayer({
     map.on('click', id, ({ features: [feature], lngLat: lng }) => {
       const coordinates = feature.geometry.coordinates.slice();
 
+      // FIXME: Add a comment for what this does
```
zoobot, line 55: // Ensure that if the map is zoomed out such that
// multiple copies of the feature are visible, the
// popup appears over the copy being pointed to.

client/src/pages/Map/MapboxManager.js
```diff
@@ -0,0 +1,85 @@
...
+    if (Array.isArray(regionType)) {
+      regionType = regionType[0];
+    }
+
+    switch (regionType) {
+      case REGION_TYPES.COUNTRY:
+        this.map.flyTo({
+          center: coords,
+          zoom: 4.5,
+        });
```
zoobot, line 39: Can we have the this.map.flyTo at the end and just have 4.5 in the switch case so you don't have to duplicate that 
```
this.map.flyTo({
          center: coords,
          zoom: 4.5,
        });
 ```
 
 Or even break out the switch case into its own function

File: client/src/pages/Map/MapboxManager.js [hunk 8]
```diff
@@ -0,0 +1,85 @@
+import { REGION_TYPES } from '../../util/constants';
+
+class MapboxManager {
```
zoobot, line 3: This looks pretty nice and clean. 
Should we use functional style paradigm with closures/factory functions to align with React dev recommends?

client/src/tests/MapboxManager.test.js
```diff
@@ -0,0 +1,75 @@
+import MapboxManager from '@/pages/Map/MapboxManager';
+import { REGION_TYPES } from '@/util/constants';
+
+describe('Test mapboxManager', () => {
```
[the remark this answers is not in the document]
zoobot: Nice tests! :)

client/src/util/constants.js
```diff
@@ -8,3 +8,31 @@ export const maintenanceActions = [
   ['brace', 'braced'],
   ['prune', 'pruned'],
 ];
+
+export const REGION_TYPES = {
```
zoobot, line 12: Awesome well placed comments here. :)

zoobot (changes requested):
LGTM when testing on local, This is a nice code cleanup.

Previously we'd added jumpTo after field testing and finding that flyTo animation is super laggy/annoying on mobile:

On mobile, we should use jumpTo(without animation). Not sure if that disappeared with some previous giant PRs but we should definitely use it/add it back in.
https://docs.mapbox.com/mapbox-gl-js/api/map/#map#jumpto

client/src/pages/Map/MapLayout.js
mwpark2014: I think the latest best practice recommendation is to not use redux except for very state-heavy front-end applications. I've read a bunch of blogs about zustand or other lightweight stores being a better alternative for apps like ours. 

That said, not sure what the tipping point would be. I think we can get away with contexts for now, but that's also because I don't want to own incorporating stores into the codebase right at this minute.

We currently have NewTree, UserLocation, and MapboxManager as contexts. Let's say we seriously consider moving to some state library after we add a couple more?

client/src/pages/Map/MapboxManager.js
mwpark2014: These are map markers, used by `MapboxMarkerPortal`. I can add a comment

client/src/tests/MapboxManager.test.js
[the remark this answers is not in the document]
mwpark2014: I think I would need a mapbox integration test to make a meaningful test for `addMarkerRefToMap`

client/src/pages/Map/MapboxManager.js
mwpark2014: What did you have in mind by "closures/factory functions"?

React components are recommended to be written with functions, but this class isn't a react component. It can be written with function closures, but that's basically what classes are in javascript anyway - syntactic sugar for stateful logic. I like the simplicity of this

mwpark2014:
> LGTM when testing on local, This is a nice code cleanup.
> 
> Previously we'd added jumpTo after field testing and finding that flyTo animation is super laggy/annoying on mobile:
> 
> On mobile, we should use jumpTo(without animation). Not sure if that disappeared with some previous giant PRs but we should definitely use it/add it back in. https://docs.mapbox.com/mapbox-gl-js/api/map/#map#jumpto

Absolutely, should be easy to add that type of logic now with this passthrough layer. 👍

client/src/pages/Map/MapboxManager.js
mwpark2014: I'm not sure if I understand. Is zoom 4.5 duplicated?

zoobot: I meant just separating the zoomLevel into its own thing to see at a glance what the zooms are.

I don't know what the default's supposed to be so not sure if it would work.
something like this with switch case if you need that default case.

```
  const zoomLevels = {
    [REGION_TYPES.COUNTRY]: 4.5,
    [REGION_TYPES.ADDRESS]: 17,
    etc: ...etc,
  };
```

or in your switch case scenario

```
  let zoomLevel = (regionType) => {
    switch (regionType) {
      case REGION_TYPES.COUNTRY:
        return 4.5;
      case REGION_TYPES.REGION:
        return 6;
      case REGION_TYPES.PLACE:
      case REGION_TYPES.DISTRICT:
      case REGION_TYPES.POSTAL_CODE:
        return 12;
      case REGION_TYPES.NEIGHBORHOOD:
      case REGION_TYPES.LOCALITY:
        return 14;
      case REGION_TYPES.ADDRESS:
      case REGION_TYPES.LATLONG:
      case REGION_TYPES.POI:
        return 17;
      default:
        return 0;
    }
  };
  ```

Comment (zoobot) [reply client/src/pages/Map/MapboxManager.js:39]: then can do a check for mobile and flyTo or jumpTo?

Comment (mwpark2014) [reply client/src/pages/Map/MapboxManager.js:39]: oh yeah, good idea!

File: client/src/pages/Map/MapboxManager.js (hunk 8 above)
Comment (zoobot) [reply client/src/pages/Map/MapboxManager.js:3]: I'll admit, I am biased towards functional paradigm and tend to avoid 'this'/classes if possible so I don't have to figure out what is 'this'
I like how much cleaner and easier to read your map Provider is making the code so am psyched about your changes and don't want to hold you up just because of this. 


Something like this avoids 'this'.

```
export const useMapbox = () => {
  return useContext(MapboxManagerContext);
};

export const MapboxManagerProvider = ({ children }) => {
  const [map, setMap] = useState(null);

  const getCenter = () => {
    if (!map) {
      return;
    }
    return map.getCenter();
  };

//other functions like setCenter, etc
  
   const mapboxManager = {
    map,
    setMap,
    getCenter,
    setCenter,
    addMarkerRefToMap,
  };

  return (
    <MapboxContext.Provider value={mapboxManager}>{children}</MapboxContext.Provider>
  );
  ```

mwpark2014: Yeah, I considered doing something similar to your example, but I am intentionally avoiding this. 

1. The first thing that's happening here is hiding a context and access to the context within the reusable `useMapbox` hook, which I know is a paradigm we use elsewhere in the code (like in `useNewTree`). I'm intentionally straying away from this, because I feel that it overcomplicates the code by creating a context and hiding it in this hook. It's not always clear what the API is (maybe typescript would help with this), so I'm forced to actually read the hook code. It takes me longer to understand what it's doing when all we're really doing is accessing a react context. I think directly accessing context within components is clearer IMO.

2. The other part is that the example rewrites `MapboxManager` as a React component, which I don't think is right. I don't think every piece of code in our front-end should be written in React as not every piece of logic is tied to rendering a UI component. Making MapboxManager think about react lifecycles overcomplicates things. 

For example, in the example you wrote, there is a pitfall embedded in the code from not using `useMemo` to memoize getCenter, setCenter, addMarkerRefToMap. On every rerender, these functions are being recreated, which is a negligible perf hit, but more importantly, this can cause breaking bugs that are hard to track down when using `debounce` or `throttle` since the functions are not static. 

Another example is https://github.com/waterthetrees/wtt_front/pull/668. It's actually quite more difficult to write this function if `MapboxManager` was a react component since we would need to use `useRef` as an escape hatch to avoid recreating the promise on each rerender. 

In general, React tries to keep components as idempotent as possible (which is why it double-renders in test envs) and uses `useEffect` hooks as an escape hatch for side effects. The problem I see when working with our map code is that there's a lot of brittle code related to our map side effects. Examples that I've identified in the past are: code that breaks down when we reduce the number of rerenders, code that is intended to be executed just once either executes on each rerender or doesn't execute at all (executes but before async requests have given it good data), which are both kinda tricky and unintuitive issues.

mwpark2014: Here's a chatgpt excerpt that actually summarizes my thoughts in a slightly different way

> In a React project, it is not necessary to make every helper function a React component. Helper functions that do not require any UI rendering should be written as normal JavaScript functions, and can be imported and used as needed in React components.

Creating unnecessary React components can add unnecessary overhead to the project and increase its complexity. Additionally, using React components for non-UI related logic can lead to confusion for other developers who may expect them to have rendering behavior.

It is generally considered best practice to separate UI-related code and business logic. UI-related code should be written in React components, while business logic should be written in regular JavaScript functions or classes.

In summary, it is not necessary to create React components for business logic helper functions that do not render anything, and it is generally best practice to separate UI-related code and business logic.

zoobot: Thank you for bringing up the not hiding contexts idea and not using useMapbox. You have me convinced. I am very on board with keeping it as simple as possible with fewer lookups for devs. 

I should have time tonight to dive in deeper into the other stuff.  Because we are heading towards larger "readability/performance" refactors of the map code section, I want to spend a bit more time trying to understand the pros and cons of class vs factory function vs custom hook vs using mapbox mapbox-gl-js out of the box for versions of MapboxManager in our app. Thanks for your patience on this PR. Also, would be nice to chat through this with you at some point.

mwpark2014:
> I just looked through PR history and jumpTo did get removed at some point in some giant PR. It is important to have jumpTo on mobile as flyTo is painful on mobile. Sorry I missed that previously. It's really not your fault it disappeared from our code base but if you can add it back in, it'll make zoom way better on mobile.

Just wanted to clarify - my initial interpretation was that `flyTo` is painful on mobile due to performance reasons. Is that true? Or is it something related to the small screen size? Because there are a couple ways to check for if a device is mobile and the one we use throughout the codebase is a screen size check. However, I think there needs to be a better "performance check" on a device to determine if we do `map.jumpTo` (and potentially remove 3d projections)

client/src/pages/Map/MapboxManager.js
mwpark2014: Okay, well, the PR itself is ready, but there were merge conflicts with main. Lmk when this is approved and I can do another round of manual testing to ensure I resolved the merge conflicts correctly - don't want to waste time on unnecessary testing cycles.

zoobot (approved):
This is a good building block and cleans up the code in a nice way. 
It's built in such a way that if we want to change MapBoxManager to a custom hook or factory function used with context later on, it wouldn't be that far off from what you have.

PR #661, opened 2023-04-17
Added @pmmmwh/react-refresh-webpack-plugin

zoobot:


PGrad:
I think it's a second or two faster, it does help. One thing: I notice that when I check out your branch and do `npm install`, there are diffs in the package-lock.json file.

PGrad (approved):
LGTM

PR #663, opened 2023-04-17
Update snaps

zoobot:
Treelist Epic:

Minor changes to the data SearchBar
Add tests for Searchbar
Update snapshots for Searchbar


PGrad (approved):
More tests are always good

PR #670, opened 2023-04-25
Fix storybook build

PGrad:
Storybook uses the 'analyze' webpack config, and some of the key-value data wasn't being pulled into the build. This submission fixes that issue so storybook will build correctly.


zoobot (approved):
LGTM

PGrad:
Thanks for the quick review!

PR #694, opened 2023-05-10
Zoobot/feature/treecare

zoobot:
This is a quick fix for https://github.com/waterthetrees/wtt_notifications/issues/22

This is a basic Treecare /How to page that will be used in the watering notifications and elsewhere.


client/src/components/App/App.js
zoobot, line 1: The only actual change in App is the TreeCare route. Sorry about the lint/import re-ordering.

client/src/components/Header/Header.js
```diff
@@ -1,10 +1,12 @@
+import { useAuth0 } from '@auth0/auth0-react';
```
zoobot, line 1: The only real change in this is the TreeCare menu item.

Import changes are all about re-ordering imports. I think someone added a rule to do that.

Issue #635, opened 2023-04-08
[Epic - User Profile] User Profile Redesign / Additions


ri0nardo:

zoobot:
We need:

1. location to store hero image per user
2. location to store image of user if it isn't pulled from gmail
3. about field for user in database - https://github.com/waterthetrees/wtt_db/blob/main/tables/public.users.sql
4.

ri0nardo:
@drumwolf @arc6789 the epic has been updated. The only ticket that has not been worked on is the U.I for the user summary section. I am pretty sure its going to be a modal, but need to explore other options. Let me know what needs more details.

PR #745, opened 2023-08-31
User Profile Redesign - refactor to eliminate styled components

drumwolf:
**ISSUE:**
https://github.com/waterthetrees/wtt_front/issues/635

**REASON:**
Currently, the `UserInfoContainer` component is using styled components, which we want to phase out.  

I've removed the styled components `UserInfoContainerBox` and `UserAvatar`, and replaced them with custom classes in a new CSS file `UserInfoContainer.scss` .

This is refactoring under the hood and should not affect the UX/UI in the browser in any way.

### DIFF
--- a/client/src/pages/Userprofile/UserInfoContainer.scss
+++ b/client/src/pages/Userprofile/UserInfoContainer.scss
@@ -2,4 +2,10 @@
   display: inline-flex;
   align-items: center;
   margin-bottom: 2em;
+
+  .user-avatar {
+    width: 8em;
+    height: 8em;
+    margin: 0 3em 0 2em;
+  }
 }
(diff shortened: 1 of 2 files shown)


client/src/pages/Userprofile/UserInfoContainer.scss
```diff
@@ -0,0 +1,11 @@
+.user-info-container {
+  display: inline-flex;
+  align-items: center;
+  margin-bottom: 2em;
+
+  .user-avatar {
```
zoobot, line 6: Here's an example of how we are nesting css so you can just have a root class
https://github.com/waterthetrees/wtt_front/blob/main/client/src/components/SourceForm/SourceForm.scss

zoobot (approved):
LGTM Thanks

